Issue #5, opened 2016-06-29
Basic testing setup


gilbrechbuhler:
- Find a way to handle xml files (with templates for example)
- Setup a basic test infrastructure for bdd
- Write some tests

PR #6, opened 2016-06-29
XML templating, BDD, first tests

gilbrechbuhler:
Handles issue #5


nigelmegitt:
This is looking pretty great - it would be good to have some documentation added to explain how to run the tests. Seems like running `make test` does something relevant but I'm not sure if I've misunderstood...

nigelmegitt:
Oh seems like I mis-guessed - it has just done a code check not run any functional or integration tests.

nigelmegitt:
Right, @kozmaz87 has explained a) that the coverage explains which lines of code were exercised and b) how to run the tests and c) what to expect if all the tests pass. I think this needs a bit better documentation including those explanations and also py.test needs to install (doesn't in my virtual environment, so I had to run `python setup.py test` manually. 

I'm basically happy with this, just needs a little bit of polishing. Thanks @gilbrechbuhler and @kozmaz87 - great to see the PRs coming in!

gilbrechbuhler:
It makes me remember that in the README.md there is :

Alternatively:

    pip install -r requirements.txt
    pip install -r requirements-test.txt
    python setup.py develop

    pyxbgen -r --binding-root=./ebu_tt_live/bindings -m __init__ --schema-root=./ebu_tt_live/xsd1.1/ -u ebutt_live.xsd

However the requirements-test.txt does not exist, should it be created or should the line be removed from the readme ?

gilbrechbuhler:
Tried this morning :
- clone repository in new folder
- create new virtualenv
- git checkout test branch
- make
- make test

And it worked perfectly, the only remaining problem is that the py.test command is not callable directly. For this to work the solution is to "pip install" pytest and the other test library (pytest-cov and pytest-bdd), so the requirements.txt file would be the solution.

Waiting on your inputs :).

gilbrechbuhler:
Otherwise is the update to the CONTRIBUTING.md file done by @kozmaz87 enough or do you want more detailed documentation ? (Like a readme in the testing/bdd folder showing the writing of a basic test ?)

nigelmegitt:
I can reproduce that: `make test` seems to work.

For consistency, it seems to me that describing how to test using e.g. `make test` should be in README.md where the instructions for building the code and the documentation already reside.

nigelmegitt:
The above commit is a minimal fix to allow for merging - we should look again at whether test documentation should be in CONTRIBUTING or README and maybe move the text across rather than just cross referencing.

nigelmegitt:
I'm happy to merge this - I understand @frans-ebu will take a look too so I'll leave open until he's able to do that, hopefully later today.

gilbrechbuhler:
Ok, and concerning the requirements-test.txt ? Should the file exist or the reference to it be removed from the README ?

nigelmegitt:
I'm not sure about this - @kozmaz87 what do you think?

kozmaz87:
I like the links with CONTRIBUTING.md I did not know how to do those. Thanks Frans

frans-ebu:
I have split of the py.test / make part into #9

frans-ebu:
@nigelmegitt I am  now also happy to merge this. Tnx to @kozmaz87 and @gilbrechbuhler

nigelmegitt:
Great, thanks all! Super team effort.

PR #10, opened 2016-07-01
Clarify Graphviz dependency

### LINKED ISSUE #8: graphviz dependency not documented
I think graphviz needs to be installed in order to see the figures in the documentation correctly, but this is not clear from the README.md.

(Building sphinx gives a warning it cannot find the dot executable if the user has not installed it).

I see two options:
- [ ] install graphviz automatically (not easy to do platform independently?)
- [x] document in README.md that the user needs to install graphviz and add the location of the dot executable to the PATH

frans-ebu:
closes #8


README.md
```diff
@@ -97,26 +97,33 @@ Documentation
...
-## Calling setuptools
+1. Calling setuptools
 
     python setup.py build_sphinx
     
-## Running make in the docs directory where separate makefiles and a make.bat file is giving a variety of options.
+2. Running make in the docs directory where separate makefiles and a make.bat file is giving a variety of options.
 
     cd docs
     make html
```
nigelmegitt: Not sure why but when I preview the file this comes out looking like one line `cd docs make html` which is misleading.

nigelmegitt:
This looks fine to me - @kozmaz87 want to take a look?

nigelmegitt:
This is fairly minor at this point, and I'd like to merge and close it. The issue, which is to document graphviz dependency, is resolved. If anyone thinks there's a pressing need to auto-install the dependency as part of the build process, please open it as a separate issue.

PR #17, opened 2016-07-03
Coveralls integration

kozmaz87:


gilbrechbuhler:
great :+1:

nigelmegitt:
Seems nifty - just want to check if there are any adverse commercial/legal implications of using coveralls, since I'm new to it?

kozmaz87:
opensource == free that is all I got from their licensing. If private there is a fee...

nigelmegitt:
@frans-ebu if you're happy with this (or don't need to check it) we can merge.

frans-ebu:
@nigelmegitt I am happy with this. I don't see a problem with it and it is not part of the code anyhow.

Issue #7, opened 2016-06-30
Timebase semantic testing


gilbrechbuhler:
Test semantic validation on timebase rules :

```
Start point of a temporal interval associated with a tt:body element.
If the timebase is "smpte" the type shall be ebuttdt:smpteTimingType .
If the timebase is "media" the type shall be ebuttdt:mediaTimingType .
If the timebase is "media" the time expression should be the offset from a syncbase of "00:00:00.0".
If the timebase is "clock" the type shall be ebuttdt:clockTimingType .
```

gilbrechbuhler:
@kozmaz87 @frans-ebu @nigelmegitt 

Next tests I will write, of course for now they will be skipped as the semantic validation is not implemented. But I think those are among the most important tests to write.

nigelmegitt:
Okay, sounds good to me. For me the important thing for us to get to is that we can validate the document activation rules in the context of a sequence of documents - spec section §2.3.1. This is a step on the way to getting there.

gilbrechbuhler:
Yeah 2.3.1 will be a big part to test, however for now, as we do not have a presentation node or anything else that actually use the timing, it is a bit hard to test. Unless I am missing something, which is possible too.

nigelmegitt:
In my mind we don't actually need to present the output to verify the timing, we just need to write out a simulation of the events that would occur during playback, i.e. a timeline with the text that would appear at each time, which gets modified as new documents become available. This can be a text output to stdout for all I care, but at least that way it is reproducible.

Also, @kozmaz87 and I have discussed the idea of having a sequence of documents stored on a file system that can be processed immediately, with a manifest file that lists a set of pairs of document (sequence number) and availability time. By using different manifest files we can test the logic for different availability time scenarios without having to update a whole bunch of EBU-TT files every time, and without having to wait for the real elapsed time to happen.

gilbrechbuhler:
I'm facing a little confusion here, from tech 3370 (Annex A. p.56) :
```
<ebuttdt:mediaTimingType>
: <full-clock-value>
| <timecount-value>

<ebuttdt:clockTimingType>
: <full-clock-value>
| <timecount-value>

<full-clock-value>
: <hh> ":" <mm> ":" <ss> ("." <digit>+)?

<timecount-value>
: <digit>+ ("." <digit>+)? <metric>

<hh> | <mm> | <ss> | <ff>
: <digit> <digit>
```
[...]

PR #18, opened 2016-07-04
Timebase semantic testing

gilbrechbuhler:
Tests that the times are accepted/rejected correctly depending on the timebase value. SMPTE is not handled yet as its implementation and usage are really likely to change in close futur.

Handles issue #7


nigelmegitt:
Great, thanks Gil, I'll take a look...

nigelmegitt:
This looks fine. Shouldn't some of the tests that are being skipped now be caught even by the pyxb binding's basic validation?

gilbrechbuhler:
Indeed like 1.5x correctly raises an exception, I thought about it and forgot to take a look, maybe it is possible to skip just some examples, I'll take a look and comment back.

gilbrechbuhler:
So commenting the scenario examples that don't work is the solution. However, as they are commented, they are not listed as skipped tests, they are just no taken into account at all. So the result about how many tests ran and how many were skipped becomes wrong.

What do you prefer ? Skipping the whole scenarios until all their members work (so the number of skipped tests is correct) or commenting the scenario examples that don't work (numver of skipped test is wrong) ?

gilbrechbuhler:
maybe @kozmaz87 as an opinion about this too

nigelmegitt:
This is fine right now but we have a choice: @gilbrechbuhler do you want to:

1. go ahead and add some more tests for `ttp:timeBase="smpte" ttp:markerMode="continuous"` following the EBU meeting on Tuesday? OR
2. merge this as is and start a new branch for smpte tests?

I don't have any semantics to test for `ttp:markerMode="discontinuous"` yet because I don't know what the spec will do with that, but for continuous see [gitlab issue 114](https://git.ebu.io/ebutt/ebutt-part-3/issues/114) where there is a requirement to test that the timing mode is the same across the whole sequence.

gilbrechbuhler:
I think I'll just quickly add tests for the check of smpte formats for now. For [gitlab issue 114](https://git.ebu.io/ebutt/ebutt-part-3/issues/114) tests I will do a new branch and a new PR.

I'll add a comment here as soon as I added the format tests for smpte here.

gilbrechbuhler:
Actually after checking, I think the best is to merge the branch as it is. Indeed if ```ttp:markerMode="continuous"``` is allowed, the XSD as to be modified in consequence (currently it only allows discontinuous), which is out of the scope of this PR. By the way, I was checking tech 3350 (for the definition of dropMode) and realized it does not allow "continuous" as a value for the markerMode, moreover with ```ttp:timeBase="smpte"``` it still says that  ```ttp:markerMode="discontinuous"``` is mandatory.

Anyway, merging this as it is, is the easier method for now as we have multiple things to change for smpte apparently.

nigelmegitt:
Okay, that's fine. The only thing remaining on this is to answer the question:

* [ ] @kozmaz87 to answer query from https://github.com/ebu/ebu-tt-live-toolkit/pull/18#issuecomment-230300247 : is there a way to selectively skip scenario examples that don't work rather than commenting them out?

frans-ebu:
I am happy with this, but admit not having checked in great depth.

nigelmegitt:
I'd like to merge this and raise an issue for skipping invididual scenario example lines that don't work rather than commenting them out. I admit I also have not checked all the tests against the spec, but any missing ones can be added later.

nigelmegitt:
The smpte tests have been moved out of #7 into #52.

The commenting out vs skipping of scenario example lines has been raised as #53.

Issue #43, opened 2016-07-05
Unit-Test documents comparison (ComparableMixin)


gilbrechbuhler:
Documents use the mixin ComparableMixin (defined in the project root in file utils.py). This mixin allows for correct and easy comparison of documents :

- two documents with the same sequenceIdentifier will be compared using their sequenceNumber : 
```
document1 < document2 if document1.sequenceNumber < document2.sequenceNumber
```

- If the documents do not have the same sequenceIdentifier, there is no comparison possible and an error is raised.

With this issue I want to address the fact that this was not tested yet, so I implemented tests to ensure that document comparison works as intended.

nigelmegitt:
Sorry, I'm not following what the issue is here - please could you add a description?

gilbrechbuhler:
Documents use the mixin ComaparableMixin (defined in the project root in file utils.py). This mixins allows for correct and easy comparison of documents :

- two documents with the same sequenceIdentifier will be compared using their sequenceNumber : 
```
document1 < document2 if document1.sequenceNumber < document2.sequenceNumber
```

- If the documents do not have the same sequence number, there is no comparison possible and an error is raised.

With this issue I want to address the fact that this was not tested yet, so I implemented tests to ensure that document comparison works as intended.

Sorry, I forgot to put a description.

nigelmegitt:
Thanks, that's clear!

gilbrechbuhler:
I think we can close this one as the corresponding PR ( #44 ) as been merged.

PR #44, opened 2016-07-05
Unit-Test comparable mixin documents

gilbrechbuhler:
For #43

### DIFF
--- a/ebu_tt_live/documents/test/test_ebutt3document.py
+++ b/ebu_tt_live/documents/test/test_ebutt3document.py
@@ -23,6 +23,6 @@ class TestEBUTT3Document(TestCase):
         self.assertRaises(ValueError, lambda: document1 < document2)
         self.assertRaises(ValueError, lambda: document2 > document1)
         self.assertRaises(ValueError, lambda: document1 <= document2)
-        self.assertRaises(ValueError, lambda: document1 >= document2)
+        self.assertRaises(ValueError, lambda: document2 >= document1)
         self.assertRaises(ValueError, lambda: document1 == document2)
         self.assertRaises(ValueError, lambda: document1 != document2)
--- a/ebu_tt_live/documents/test/test_ebutt3document.py
+++ b/ebu_tt_live/documents/test/test_ebutt3document.py
@@ -1,0 +1,26 @@
+from unittest import TestCase
+from ebu_tt_live.documents import EBUTT3Document
+
+
+class TestEBUTT3Document(TestCase):
+
+    def test_comparison_same_sequence(self):
+        document1 = EBUTT3Document("clock", 1, "testSeq", "en-GB", "local")
+        document2 = EBUTT3Document("clock", 2, "testSeq", "en-GB", "local")
+        self.assertTrue(document1 < document2)
+        self.assertTrue(document2 > document1)
+        self.assertTrue(document1 == document1)
+        self.assertTrue(document1 <= document2)
+        self.assertTrue(document2 >= document1)
+        self.assertTrue(document1 <= document1)
+        self.assertTrue(document1 >= document1)
+
+    def test_comparison_different_sequences(self):
+        document1 = EBUTT3Document("clock", 1, "testSeq1", "en-GB", "local")
+        document2 = EBUTT3Document("clock", 2, "testSeq2", "en-GB", "local")
+        # assertRaises does not catch the error correctly without the lambda.
+        self.assertRaises(ValueError, lambda: document1 < document2)
+        self.assertRaises(ValueError, lambda: document1 > document2)
+        self.assertRaises(ValueError, lambda: document1 <= document2)
+        self.assertRaises(ValueError, lambda: document1 >= document2)
+        self.assertRaises(ValueError, lambda: document1 == document2)
--- a/ebu_tt_live/documents/test/test_ebutt3document.py
+++ b/ebu_tt_live/documents/test/test_ebutt3document.py
@@ -14,6 +14,7 @@ class TestEBUTT3Document(TestCase):
         self.assertTrue(document2 >= document1)
         self.assertTrue(document1 <= document1)
         self.assertTrue(document1 >= document1)
+        self.assertTrue(document1 != document2)
 
     def test_comparison_different_sequences(self):
         document1 = EBUTT3Document("clock", 1, "testSeq1", "en-GB", "local")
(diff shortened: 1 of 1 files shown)


ebu_tt_live/documents/test/test_ebutt3document.py
```diff
@@ -0,0 +1,28 @@
...
+        self.assertTrue(document1 <= document1)
+        self.assertTrue(document1 >= document1)
+        self.assertTrue(document1 != document2)
+
+    def test_comparison_different_sequences(self):
+        document1 = EBUTT3Document("clock", 1, "testSeq1", "en-GB", "local")
+        document2 = EBUTT3Document("clock", 2, "testSeq2", "en-GB", "local")
+        # assertRaises does not catch the error correctly without the lambda.
+        self.assertRaises(ValueError, lambda: document1 < document2)
+        self.assertRaises(ValueError, lambda: document1 > document2)
```
nigelmegitt: Should this be document2 > document1 ? They're reversed in the assertTrue statements... Similarly below for >=.

ebu_tt_live/documents/test/test_ebutt3document.py
```diff
@@ -0,0 +1,28 @@
...
+        document1 = EBUTT3Document("clock", 1, "testSeq", "en-GB", "local")
+        document2 = EBUTT3Document("clock", 2, "testSeq", "en-GB", "local")
+        self.assertTrue(document1 < document2)
+        self.assertTrue(document2 > document1)
+        self.assertTrue(document1 == document1)
+        self.assertTrue(document1 <= document2)
+        self.assertTrue(document2 >= document1)
+        self.assertTrue(document1 <= document1)
+        self.assertTrue(document1 >= document1)
+        self.assertTrue(document1 != document2)
```
nigelmegitt: Worth adding document2 != document1 ?

ebu_tt_live/documents/test/test_ebutt3document.py
gilbrechbuhler: Indeed, thanks for pointing out the inconsistency

ebu_tt_live/documents/test/test_ebutt3document.py
gilbrechbuhler: It was just for the sake of testing each comparison operator, but indeed as two documents cannot have the same sequence ID and the same sequence number, this comparison will mostly never be used. If you think it is irrelevant to test that case, I can remove this line.

nigelmegitt: It's possible to construct two document objects that have the same sequence id and number, so I think it is relevant to test - my point was just that we need to test it in both directions.

gilbrechbuhler: Oh right, my bad, sorry, I misread your comment, I will add it.

nigelmegitt: Ok, I thought you might have been checking that the operators work correctly independent of the order of the operands too. If (document1 != document2)==TRUE but (document2 != document1)==FALSE then we have a big problem!

gilbrechbuhler: Indeed, but I don't see how it could happen due to the semantic of the ComparableMixin. But I'll add the test, you are right it is safer.

nigelmegitt:
Thanks, looks good to me!

frans-ebu:
Looks good.

On the double sided inequality check:
- [ ] What would be an example of this going wrong?
- [ ] If we feel it has value, why then not apply it to `test_comparison_different_sequences(self)` then as well?

Before merging we should understand why Travis chokes in the /pull/49 part...

gilbrechbuhler:
When we compare different sequence, the code does not even go to the comparison part, it raises an error before. This is why I did not add double sided tests for this part, the code takes the two documents and compare their sequence identifier, so the side of the inequality does not change anything. But if you want me to add double sided checks, for safety, I can do it.

For travis choking I'll take a look, it is weird.

gilbrechbuhler:
Ok so travis tries to checkout merge request #49  but this one is #44. This is really strange, why would he do that ? Of course he cannot checkout PR #49 as it has been merged and does not exist anymore. I don't understand what happens. Maybe it is a bit disturbed by the fact that you created another merge request on the exact same branch to test notifications, I'm not sure of what I say here, I'm just trying to guess the issue. But I'm pretty sure the problem is git/github related here and that everything will be right on other PRs.

gilbrechbuhler:
Ok now I'm starting to be pretty sure I'm right. It started to fail right after merging #44 in notification-test. Times are corresponding.

frans-ebu:
yes, I thought that, too
seems a bug inTravis?!

F

Sent from my iPhone

On 06.07.2016, at 21:26, Brechb�hler Gil <notifications@github.com<mailto:notifications@github.com>> wrote:


Ok now I'm starting to be pretty sure I'm right. It started to fail right after merging #44<https://github.com/ebu/ebu-tt-live-toolkit/pull/44> in notification-test. Times are corresponding.

�
You are receiving this because you were assigned.
Reply to this email directly, view it on GitHub<https://github.com/ebu/ebu-tt-live-toolkit/pull/44#issuecomment-230880132>, or mute the thread<https://github.com/notifications/unsubscribe/AEXZTVve-51ZjQwHH5-GR_bjtTuQF2A3ks5qTAFQgaJpZM4JE8-X>.
------------------------------------------------------------------------------

**************************************************
This email and any files transmitted with it are confidential and intended solely for the use of the individual or entity to whom they are addressed.
If you have received this email in error, please notify the system manager. This footnote also confirms that this email message has been swept by the mailgateway
**************************************************

frans-ebu:
Regarding the (in)equality testing, I guess I agree with Nigel that theoretically this could go wrong (e.g. if the code would have implemented something like return doc1.somebooleanthatistrue), but that is highly unlikely as we know the code.

But then the same case can be made for any == or != operation we define, so we should add it to the other 3 cases in the same file, too?

self.assertTrue(document1 == document1)
self.assertRaises(ValueError, lambda: document1 == document2)
self.assertRaises(ValueError, lambda: document1 != document2)

gilbrechbuhler:
I will add them

gilbrechbuhler:
Well Travis decided to work finally, it was certainly a bug.

frans-ebu:
Tnx, ready to merge!

Issue #51, opened 2016-07-08
Create text equivalent list of normative spec requirements


nigelmegitt:
Generate a text format list of the normative spec requirements against which we can write tests.

nigelmegitt:
This will help with #45.

nigelmegitt:
Working on this at https://github.com/ebu/ebu-tt-live-toolkit/blob/issue-51-conformance-requirements/testing/SPEC-CONFORMANCE.md if anyone wants to take an early look.

frans-ebu:
One thing I wondered is if we want to add the origin of the requirement. E.g. in an additional column?
Now all are from Tech 3350, but as you said, there may be others from related specs (Part 1, TTML1SE, etc.).

nigelmegitt:
Good idea!

PR #54, opened 2016-07-08
Issue 51 conformance requirements

nigelmegitt:
Fix issue #51 by adding a conformance requirements list markdown document. Needs tests to be added to the right hand column, and also possibly conformance requirements referenced by Tech3370.

### DIFF
--- a/testing/SPEC-CONFORMANCE.md
+++ b/testing/SPEC-CONFORMANCE.md
@@ -1,4 +1,4 @@
-# Conformance Requirements list for EBU-TT Part 3 Tech3370
+# Conformance Requirements list for EBU-TT Part 3 Tech 3370
 
 The conformance requirements for EBU-TT Part 3 derive from the specification itself, EBU-TT Part 1, TTML1SE and their referenced specifications - SMIL, XSL-FO etc. The following normative statements are present in Tech3370:
 
--- a/testing/SPEC-CONFORMANCE.md
+++ b/testing/SPEC-CONFORMANCE.md
@@ -4,7 +4,7 @@ The conformance requirements for EBU-TT Part 3 derive from the specification its
 
 ## SHALL requirements
 
-|ID|Section|Statement|Test|
+|ID|Section|Statement|Scenario name|
 |---|---|---|---|---|
 | | **Tech3370**| | |
 | R1|2.2|When presenting a sequence of documents, at each moment in time exactly zero or one document shall be active.| |
--- a/testing/SPEC-CONFORMANCE.md
+++ b/testing/SPEC-CONFORMANCE.md
@@ -1,0 +1,78 @@
+# Conformance Requirements list for EBU-TT Part 3 Tech3370
+
+The conformance requirements for EBU-TT Part 3 derive from the specification itself, EBU-TT Part 1, TTML1SE and their referenced specifications - SMIL, XSL-FO etc. The following normative statements are present in Tech3370:
+
+## SHALL requirements
+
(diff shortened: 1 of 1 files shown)


frans-ebu:
Looking good!

I have added one proposal in 0b53ac4 : to make section references explicit, so rows can be read in isolation (may be handy for future auto-ref listing, as well).

gilbrechbuhler:
Great, I'll think of way of referencing the tests in a readable and understandable way.

nigelmegitt:
Thanks @frans-ebu I wondered about that, happy to go with the ref-on-every-line approach.

gilbrechbuhler:
I think the easiest way to reference the tests is to give the path to the file containing the test (feature file in case of bdd) from the root of the project. If you have better ideas :)

nigelmegitt:
What about the scenarios within the feature file?

gilbrechbuhler:
Yes for feature file that might be a great idea. However, I suspect some of the tests for conformance requirements may belong in unit tests, in that case we could just put the test file path's ?

gilbrechbuhler:
Looking at the file, a question came to me :
`Every distinct document with the same sequence identifier shall have a different sequence number.`

How do we plan on enforcing these kind of statements ? I mean, either we keep a list of each sequence going through each node and check on every node, or we simply guarantee that the producers we provide never break this rule and test them ?

I ask because the first solution seems to not be the right one (big performance problems could come from this I think). But in the case on the second one, if someone creates its own producer, we cannot really enforce that, we can only enforce it in our code, the consequence if this error is made will be that the consumer will not have the expected behaviour.

So how should this kind of statements be tested ? Should I just test our producers to be sure that they do not break the rule for example ?

nigelmegitt:
It's a semantic validation test - if you receive a document with a sequence number that's been seen before, and the purpose of the script is to validate the whole sequence, then you have to check that the _n_ documents with the same sequence number are identical. We're testing conformance here, not trying to be performant.

gilbrechbuhler:
Ok so testing this means testing the code/node that performs semantic validation in that case, right ?

For perfomances, I thought validating this on every node, but if it is only on a node that performs semantic validation, this is not an issue indeed.

nigelmegitt:
Yes, I think that's right.

gilbrechbuhler:
May I complete the file in this PR ? Some tests already exist.

nigelmegitt:
Yes, sure. I'm not sure what you're going to add?

nigelmegitt:
Should we be extra clear and change the Tests column header to be "Scenario name"?

gilbrechbuhler:
I think it is a good idea, for now we only reference scenarios, if that changes one day we can always update the column name.

nigelmegitt:
Ok, agreed. Please could you make that tweak too?

nigelmegitt:
Looking good, thanks. Is there something in the test framework that makes scenario names unique, or should we also be including some kind of test path and/or feature name or id?

gilbrechbuhler:
Arf, no, as long as they are not in the same `.feature` files, scenarios can have the same name. Instead of the full path, should I add the `.feature` filename in front of each scenario ?

nigelmegitt:
Yes, `path/to/some.feature` `Scenario`

nigelmegitt:
I think this is looking good. @kozmaz87 does it do enough to answer your question about how we measure how much of the spec we are testing?

kozmaz87:
This is excellent. Happy to go with it.

Issue #53, opened 2016-07-08
skip individual scenario example lines


nigelmegitt:
It would be good to be able to skip individual example lines in scenarios without commenting them out, so that those scenarios are counted better.

gilbrechbuhler:
I thought that something like :

```gherkin
Scenario: Invalid times according to timeBase
    Given an xml file <xml_file>
    And it has timeBase <time_base>
    And it has body begin time <body_begin>
    And it has body end time <body_end>
    And it has body duration <body_dur>
    Then document is invalid

    Examples:
    | xml_file                 | time_base | body_begin    | body_end      | body_dur      |
    | timeBase_timeformat.xml  | clock     | 15.58a        | 1.5d          |               |
    | timeBase_timeformat.xml  | media     | 67.945q       | 125.0x        |               |
    | timeBase_timeformat.xml  | media     |               |               | 99.9l         |
    | timeBase_timeformat.xml  | media     | 42:05:08:60.8 | 45:00:47.0    |               |
    | timeBase_timeformat.xml  | media     | 140:09:60.8.1 | 141:00:60.999 |               |
    | timeBase_timeformat.xml  | media     |               |               | 225:59:60.9.3 |
    @skip
    | timeBase_timeformat.xml  | clock     | 0142:05:60.8  | 145:00:47     |               |
    | timeBase_timeformat.xml  | clock     |               |               | 199:00:60.4   |

```

So my idea was to group all tests that do cannot pass at the end of the examples and skip them only. It works, in the sense that those tests are not run. However they are not counted either.
I think that when pytest parses the table, it stops at the first line that does not start with a comment or "|", or something like that.

I'm still looking for a solution but I am not sure this is possible.

nigelmegitt:
That's still more elegant than comments, even if the counting semantics aren't ideal.

PR #58, opened 2016-07-11
Skip examples

gilbrechbuhler:
For issue #53

### DIFF
--- a/testing/bdd/features/validation/timeBase_timeformat_constraints.feature
+++ b/testing/bdd/features/validation/timeBase_timeformat_constraints.feature
@@ -21,7 +21,6 @@ Feature: ttp:timeBase-related attribute constraints
 
 
     # These tests are not all passing because the missing semantic validation piece
-  #@skip
   Scenario: Invalid times according to timeBase
     Given an xml file <xml_file>
     And it has timeBase <time_base>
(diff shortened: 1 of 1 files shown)


gilbrechbuhler:
We simply group at the end the examples that cannot pass and "skip" them. However they do not appear in the "skipped" count.

nigelmegitt:
Does the skip count normally include a count of every row in the example table in the scenario, or just count each skipped scenario? If it's per-scenario then it's going to be difficult to get this to work 'perfectly'.

gilbrechbuhler:
Normally, when we use @skip on a scenario each row is counted. Here when we skip just some lines they are not counted, I bet it comes from the way pytest parses the tables

nigelmegitt:
Yes, seems like a problem with pytest. I suggest we use the closest we can to a sensible format, i.e. putting `@skip` part way down the list of examples, and raise an issue with pytest.

gilbrechbuhler:
Agreed. However I'm not sure the issue is specific to pytest, I searched for means of commenting scenario examples in other bdd framework and I did not find anything interesting sadly.

nigelmegitt:
Ok, maybe it's a cucumber/gherkin problem then - either way it seems like something we should escalate to wherever it needs to go!

gilbrechbuhler:
Merge-able for me, no conflict.

frans-ebu:
Let's merge :)

nigelmegitt:
There seems to be consensus to merge this: doing so.

Issue #50, opened 2016-07-07
File system carriage mechanism


gilbrechbuhler:
Refers to [BBC's fork issue 18](https://github.com/bbc/ebu-tt-live-toolkit/issues/18)

This issue asks for the implementation of a functionality that allows the simple producer to write its output to the file system along with a manifest file with availability times.

nigelmegitt:
This is effectively a 'file system' based carriage mechanism. Will be very useful for testing and debugging processes if nothing else.

gilbrechbuhler:
Thank you for pointing the misnaming here, corrected.

gilbrechbuhler:
Ok, the local clock used by the SimpleProducer is a local clock which has :

```
def get_time(self):
        now = datetime.now().time()
        return timedelta(hours=now.hour, minutes=now.minute, seconds=now.second, microseconds=now.microsecond)
```

Which means that we never get the date included. And I use the producer's clock to get "available times". However, @frans-ebu pointed out that question : "how do we handle the case when we have a manifest file that spans over two days if we do not have dates ?".

I ask this question here because we have to use the node's clock to make sure the "availability" times are coherent with the times in the produced document.

nigelmegitt:
We need to put times in the manifest file that can be compared with document times, so we should either put the availability times in the same timeBase as the documents, or provide some kind of offset value to show how to relate the manifest times with the document times. Agreed during call of 2016-07-13 to use the same timeBase because it's simpler.

gilbrechbuhler:
As currently the code that writes the manifest file uses the producer's clock, it is already the case. I need to see if it works well with a media clock.

gilbrechbuhler:
Let's break down what is left to do :

- [ ] Add the availability time as an attribute of `EBUTT3Document` class
- [ ] Write a consumer node able to produce a text output simulating the documents and there activation times

PR #59, opened 2016-07-11
Simple producer filesystem export

gilbrechbuhler:
Quite a lot of changes here. First, to test it, simply launch :
```
ebu-simple-producer --folder-export fs_output
```
You can name the folder as you want, it will be created in your current directory. It contains one xml file per document and a **manifest.txt** file that has the following format :
```
availability_time,filename 
```
We can discuss the format. Note that the manifest file is written using the producer-node's clock in order for the availability times to be coherent with the timing of the documents.

I had to add a "carriage" package, as this is a new carriage mechanism, and it clearly did not have his place in "twisted" package. Also, I moved twisted's carriage implementations in the carriage package and left the rest, that is using twisted more directly, in the twisted package. We might need to discuss about code organization here. I talked with @kozmaz87 quickly, we agreed that we cannot put the whole "twisted" package in the "carriage" package as it does more (event loop, ...).

Now I'm wondering how to use it efficiently for futur tests.

For issue #50


gilbrechbuhler:
working on basic unittest + a correction now

gilbrechbuhler:
The correction is that before the manifest file descriptor was open during the whole lifetime of an instance, that was not good. Now the manifest content is cumulated in a str variable and wrote to the manifest file in one go at the end.

nigelmegitt:
Great, thanks for this. I will look at it but it'll take me a few days. Everyone else can have a look and add their review comments of course.

gilbrechbuhler:
Reviewable and merge-able. Please give your opinion on the new file organization for the support of several carriage mechanisms

ebu_tt_live/carriage/__init__.py
```diff
@@ -0,0 +1,2 @@
+from filesystem import FileSystemCarriageImpl
```
nigelmegitt: Do we now have a single init file that imports everything for all carriage mechanisms even if they are not actually used? Seems undesirable.

frans-ebu:
A small comment on the manifest.txt style. I noticed that if the ms part is 0 it is not shown at all. As all other times have 6 ms digits, this could potentially confuse? 

```
18:59:15.993000,TestSequence1_133.xml
18:59:16,TestSequence1_134.xml
18:59:16.008000,TestSequence1_135.xml
```

gilbrechbuhler:
Oh I did not spot that (never happened to me to have 0ms when I tested), working on a solution to always have 6 ms digits

frans-ebu:
Tnx, one more style question: the format of the start time (ms) and the duration (hh:mm:ss format) are different, would it not be more logical to have the same format? (One could argue having different ones shows what could happen, but I think we'd better show something intuitive.)

nigelmegitt:
Seems like we need to document the manifest file format - perhaps in the wiki here, and also in code documentation.

nigelmegitt:
To recap on a conversation from yesterday, we can't successfully complete review of this until we also have something that can receive and process documents from the file system via a manifest file.

gilbrechbuhler:
Quite a progress on this one, I added a simple consumer carriage part that does not do anything with the availability times, we will have to see how we want to use them in our tests.

The consumer carriage implementation only defines `on_new_data(self, data)` method, and it has to be triggered manually so, in order to not change the interface `ConsumerCarriageImpl`, I needed the class `FilesystemReader` which actually reads the manifest line by line and sends the data to the ConsumerCarriageImpl that can then handle the documents and availability times. Here it just send the document to the node and discards the availability time, but it is normal as the usage of the availability time will be case-specific I think.

@kozmaz87 I'd need your opinion about all that and any remarks that you have, thank you.

gilbrechbuhler:
Oh and for #64  here is the documentation I wrote :

```
This class implements a carriage mechanism to output produced documents
    to the file system. Its constructor takes a mandatory argument : the path to
    the desired output folder. If the folder does not exist it will be created.
    Each document handled by this carriage implementation is written in an xml file
    in the output folder.
    Along with the xml files, a manifest.txt file is also written in the output folder.
    Each time an xml file is written, a line using the following format is added to the manifest :
    `availability_time,path_to_xml_file`
    The manifest file gives the availability time for each document along with the path to the
    corresponding document.
    The timeline used for the availability times is the same as the one used in the documents,
    indeed the carriage implementation uses the same clock (or time reference) as the node that
    produces the documents.
    The writing order and thus the reading order is from top to bottom. Please note that depending
    on the clock used by the producer node, time may loop (going to the next day), this is especially
    the case with `ttp:timeBase="clock"`.```

**Comment by nigelmegitt** [comment]
It looks like we need to refactor something in the consumerImpl if we have to discard availability times right now - processing with respect to availability times is a must have, since we want to use the interop toolkit to verify behaviour according to the spec, and the spec requires comparison with availability times. As it stands I'm expecting the filesystem carriage spec to simulate processing using the availability times in the manifest for this logic.

File: ebu_tt_live/carriage/filesystem.py [hunk 2]
```diff
@@ -0,0 +1,126 @@
...
+    `availability_time,path_to_xml_file`
+
+    The manifest file gives the availability time for each document along with the path to the
+    corresponding document.
+    The timeline used for the availability times is the same as the one used in the documents,
+    indeed the carriage implementation uses the same clock (or time reference) as the node that
+    produces the documents.
+    The writing order and thus the reading order is from top to bottom. Please note that depending
+    on the clock used by the producer node, time may loop (going to the next day), this is especially
+    the case with `ttp:timeBase="clock"`.
```
nigelmegitt: Also `ttp:timeBase="smpte"`, and it is _not_ the case with `ttp:timeBase="media"`.

gilbrechbuhler:
Euh, availability times are completely usable in the FilesystemConsumerImpl, we do not *have to* discard them, they are here and usable by the CarriageImpl, I just don't know how we want to use them. So they are there, *usable*, we can add any logic we want, like having a clock running and sending documents each time an availability time is reached. But a lot of questions arise here, how do we simulate the clock ? What logic do we want ? etc...

nigelmegitt:
Here's how I thought we'd use the availability times and document times: Load each document in sequence, ordered by ascending availability time, emit a semantic validation output for each document (and the sequence) as it is understood at its availability time, then move on to the next one without waiting. That way a whole sequence can be processed 'instantly' with the same simulated results as if they were being generated in real time according to the manifest. 

For example say we have:

- Seq. num 1 availability time 15:00:00, earliest computed begin time = 15:00:05, latest computed end time = 15:00:15, presented text is "first document"
- Seq. num 2 availability time 15:00:10, earliest computed begin time = 15:00:20, latest computed end time = 15:00:30, presented text is "second document"

Load doc 1:
the view now is that we have:
```
Document sequence number 1 loaded, availability time 15:00:00.
ISDs:
15:00:00-15:00:05 - no text presented
15:00:05-15:00:15 - "first document"
```

Load doc 2:
updated view is now that we have:
```
Document sequence number 2 loaded, availability time 15:00:10.
ISDs:
15:00:00-15:00:05 - no text presented
15:00:05-15:00:10 - "first document"
15:00:10-15:00:20 - no text presented
15:00:20-15:00:30 - "second document"
```

etc as each document is loaded in turn. There could be a shortcut text output that says on loading 2nd document something like:

```
Document sequence number 2 loaded, availability time 15:00:10.
Document sequence number 1 truncated, now ends at: 15:00:10.
ISDs:
15:00:10-15:00:20 - no text presented
15:00:20-15:00:30 - "second document"
```

or something like that. Does that help explain what I mean?

kozmaz87:
About the carriage package. I can see you are exposing things there such as the twisted subpackage. I would rather that the twisted subpackage would be referenced in an import rather than the carriage package always importing the twisted bits.

So instead of:

    from ebu_tt_live.carriage import TwistedConsumerImpl

this is better:

    from ebu_tt_live.carriage.twisted import TwistedConsumerImpl

This does not require the following line in the carriage __init__.py

    from twisted import TwistedConsumerImpl, TwistedProducerImpl

This way no unused import warnings and greater flexibility. Let's assume that the twisted framework is not present the filesystemexport should still be usable.

gilbrechbuhler:
@nigelmegitt Oh, I imagined something far more complicated, sorry for the misunderstanding. I thought we wanted to have a simulated clock and do tests in real time. Ok so what you describe would be done in a specific consumer node made for that, that changes some things and it makes me realize that I did not approach the FilesystemConsumerCarriage the right way, it can be improved, FilesystemReader might not be needed.

@kozmaz87 thanks for the remark, you are right, I will correct that.

gilbrechbuhler:
@nigelmegitt The simulation you describe is basically a simulation of a presentation node if I'm getting you right. Implementing that might take some time if you want the full simulation along with this pull request though, do you mind if I keep that in my backlog but focus first on tests ?

@frans-ebu

ebu_tt_live/carriage/filesystem.py
```diff
@@ -0,0 +1,131 @@
...
+            self._manifest_time_format = MANIFEST_TIME_CLOCK_FORMAT
+        filename = '{}_{}.xml'.format(document.sequence_identifier, document.sequence_number)
+        filepath = os.path.join(self._dirpath, filename)
+        with open(filepath, 'w') as f:
+            f.write(document.get_xml())
+        # To be able to format the output we need a datetime.time object and
+        # not a datetime.timedelta. The next line serves as a converter (adding
+        # a time with a timedelta gives a time)
+        time = (datetime.datetime.min + self._node.reference_clock.get_time()).time()
+        self._manifest_content += '{},{}\n'.format(time.strftime(self._manifest_time_format), filename)
```
nigelmegitt: It would make sense to append to the manifest file as well as the in-memory manifest at this point. I see that you can `open(manifest_path, 'a')` for this purpose. Otherwise the manifest is only written at the end of `resume_producing` and then any reader never sees anything until everything has completed.

gilbrechbuhler: Indeed, when I wrote this, I never thought about the possibility to read the manifest file as it is written. I will change that.

gilbrechbuhler:
Just to be sure, I have to say that I cannot continue this one without the "fixed time mode" for the clock ( #81 ). It was made for that functionnality at first.

nigelmegitt:
Just realised I wrote this previously and somehow didn't post the comment:

No, you definitely need a FileSystemReader or something similar, to provide documents from the file system to the ConsumerNode. I've possibly changed my mind in the interests of making progress, and am now happy to factor that out into a separate issue and say this is done for now, just to write files and a manifest to the file system.

Also, as discussed off github, we will add an availability time property to `EBUTT3Document` in a separate PR and use that for downstream semantic processing, which removes the dependency on #81.

nigelmegitt:
See #102 also.

gilbrechbuhler:
We can do that what I can do is factor out the consumer part on a local branch, so we keep only the producer part here and raise another issue, would you be okay with that ?

nigelmegitt:
Yes, happy to factor out the reader part and make it a separate branch/PR for an associated issue.

kozmaz87:
I am happy with that too.

gilbrechbuhler:
Ok so I think the producer part is in a good state, I tested the existing code (simple producer and simple consumser) with the refactoring implied by the `carriage` package, everything works well

kozmaz87:
ACK. Go.

nigelmegitt:
Note sure if this is just a documentation note that's needed, but running `ebu-simple-producer --folder-export test_export` twice consecutively now creates new documents with duplicate sequence numbers and different availability times, so you get something like:

```
15:51:13.139144,TestSequence1_172.xml
15:51:13.151262,TestSequence1_173.xml
15:51:13.158268,TestSequence1_174.xml
15:51:13.164416,TestSequence1_175.xml
15:51:13.170925,TestSequence1_176.xml
15:51:13.176698,TestSequence1_177.xml
15:51:54.205768,TestSequence1_1.xml
15:51:54.212020,TestSequence1_2.xml
15:51:54.218221,TestSequence1_3.xml
15:51:54.224721,TestSequence1_4.xml
15:51:54.231053,TestSequence1_5.xml

...

15:51:55.320740,TestSequence1_174.xml
15:51:55.326873,TestSequence1_175.xml
15:51:55.332785,TestSequence1_176.xml
15:51:55.338797,TestSequence1_177.xml
```

Three possibilities (there may be more):

1. Do nothing to code; add a usage note.
2. Detect an existing manifest file on startup, then exit or clear it.
3. Detect an existing manifest file, parse the last line, start sequence numbers from a higher number.

Thoughts?

gilbrechbuhler:
Oh thanks for spotting that, I should have seen the problem.
Solution 1 is not the good one I think.
Solution 2 where we clear the manifest file is the most intuitive for me. 
But 3 is good as well, not harder to implement and may be useful in some cases (I'm guessing here) but it is less intuitive in my opinion. If we chose 3 I will add documentation explaining the behavior.

Any other thoughts ?

nigelmegitt:
Well 2 actually adds a further problem - what to do if there's a file name clash with existing XML files with the same sequence numbers? So 3 is probably best, overall.

gilbrechbuhler:
Solved with solution n°3. You might wonder why I do the check in the `resume_producing` method. That is because I need to access the document sequence that is defined in the node. And the registers itself in the carriage implementation in its constructor. So in the carriage implementation constructor I don't have any node registered yet. This was the best solution I could think about. Tell me if it is ok for you.

nigelmegitt:
You mean you've put it in `resume_producing()` because `init()` is too soon? Seems very reasonable to me. All looking good for me.

gilbrechbuhler:
It is exactly what I meant. Perfect then.

gilbrechbuhler:
This time I think it is finally done. I'm going to work on the consumer part on the new branch now

nigelmegitt:
Looking good - as far as I'm concerned this can be merged.

kozmaz87:
ACK. This is all good.

Issue #67, opened 2016-07-14
TimecountTimingType regex is wrong


gilbrechbuhler:
Spotted this while working on my understanding of the bindings and how to add smpte <-> timedelta conversion. 

Python's regexes are read sequentially, meaning that if you have for example `[0-9]+(h|m|s|ms)` :
- 9h is parsed normally
- 9m also
- 9s also
- However, 9ms is parsed as 9m and the s is forgotten

This is really problematic because it also allows 9mh for example and extracts 9m, however 9mh is not permitted in this case.

To solve the problem, the solution is to add a `$` at the end of the regex, so for timecountTimingtype : 
- `[0-9]+(\.[0-9]+)?(h|m|s|ms)$` in the xsd
- `?P<numerator>[0-9]+(?:\\.[0-9]+)?)(?P<unit>h|m|s|ms)$` in bindings

gilbrechbuhler:
Okay so XML begin... what is is. ^ and $ cannot be used in regex, so we cannot use them when we define our types in XSD files. The solution here is then : `[0-9]+(\.[0-9]+)?(h|ms|s|m)` (reverse m and ms in the regex so that ms is parsed first)

PR #68, opened 2016-07-14
fix timecountTimingType bug

gilbrechbuhler:
Okay I said I would open less merge requests, but this fix was so quick to write and does not interfere with any other PR.

Look at #67 for explanations.

@kozmaz87 do you agree with the fix ?

### DIFF
--- a/ebu_tt_live/bindings/_ebuttdt.py
+++ b/ebu_tt_live/bindings/_ebuttdt.py
@@ -42,7 +42,7 @@ class TimecountTimingType(_TimedeltaBindingMixin, ebuttdt_raw.timecountTimingTyp
     """
 
     # NOTE: Update this regex should the spec change about this type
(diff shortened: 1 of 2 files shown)


gilbrechbuhler:
Ok so XML and XSD do not support ^ and $ in their regexes

kozmaz87:
I'm happy with it.

nigelmegitt:
Are there accompanying tests that verify this works? Now that you've found the fault scenario we should add it to tests to make sure it doesn't get reintroduced by some well meaning person later (someone like me who comes around and "tidies up" without realising I'm breaking it).

frans-ebu:
Elegant solution. I'd say add an issue for the tests @nigelmegitt suggested and merge this one.

nigelmegitt:
okay I'm happy to do that.

nigelmegitt:
Issue is #69 as just referenced. Will merge...

nigelmegitt:
Thank you @gilbrechbuhler - another great PR!

Issue #80, opened 2016-07-20
Create a "stoppedClock" (or similar name) clock


gilbrechbuhler:
We need a clock on which we can do : `clock.set_time(time)` that will always return `time` if we do `clock.get_time()` unless we change `time` with another call to `set_time`

nigelmegitt:
This would support something like:

As a _tester_ I want a node to process a sequence with pre-defined per-document availability times so that the test output is consistent regardless of when the test is run.

kozmaz87:
get_time() function should be overwritten in this case to serve the saved value... unless there is some crazy usecase which requires the normal functionality along with the stopped clock functionality at the same time....

PR #81, opened 2016-07-21
First try to be able to save time in a clock

gilbrechbuhler:
This is the easiest solution I could find after some thinking. The advantage of this solution is that it allows any class inheriting from Clock to work in stopped time mode.
 Please tell me what you think about this solution and any remark you have.
For issue #80


nigelmegitt:
So this means that the code using the clock has to know whether to call `get_time()` or `saved_time()`? Wouldn't it be cleaner to make this decision a part of the clock's setup so that `get_time()` is the only function that needs to be called? I'm just wondering about the impact of this proposal on the calling code.

gilbrechbuhler:
Mmmh, actually I don't know what is the best here. The logic I had in mind was that it is the calling code that decides if it wants to get the real time of the clock or the saved_time. As the choice to use one or the other is completely context dependant.

By using the same clock, the calling code can either get the real time of the clock with `get_time()` or get the last saved_time with `clock.saved_time`.

But it may not be the best solution.

gilbrechbuhler:
Do you want me to think on a clean solution in which the decision would be made during the clock's setup and propose it on another branch ?

The thing is, I'd like to avoid refactoring all the clock classes that already exist, this is why I did not go this way, but I can try.

nigelmegitt:
Sounds like a good idea, yes. I've just been looking at where the clock is used and it seems to be on the sequence. The thing I'm expecting to see, that's missing, is any availability time property on `EBUTT3Document`, which is strange - I don't understand how the document can be processed without this data. It seems to me that the carriage mechanism code that provides documents to the `ConsumerImpl` needs to be creating `EBUTT3Document` objects which know their availability time. I wonder if @kozmaz87 had plans to add this at some point...

gilbrechbuhler:
But as you can see get_time is implemented in Clock's class childs, not in the base class, so it is a class dependant function. That is why I wanted to simply create a new class for this clock at the beginning. Then the choice is made during the clock setup, when you set the class of the clock like `myClock = AnyClock()`

gilbrechbuhler:
Concerning the availability time "location" I don't know, indeed did you have a specific plan for that @kozmaz87 or will it just be a property of `EBUTT3Document` objects ?

kozmaz87:
My opinion is the following: get_time should be the one this test clock should use as well. You do not want to if-else everywhere where clocks are ever used. Remember, most of the functionality that makes use of the realtime/media clocks are not on master yet. So Nigel I am with you on this one. get_time should just give back the set value in this case.

About the availability time
Indeed... well the availability time is a system-pipelineing concept. The activation and deactivation times are document related concepts. So in the producer side of things availability time is almost not even meaningful since availability time is the time from which the document is available to the consumer. For producer I just took a time 1 second earlier than the document activation time and hardcoded that. We can obviously work on these. 

On the consumer side of things attaching the availability time to the EBUTT3Document makes sense. We can certainly do that. However once the consumer scanned the timings in the document and worked out that it is not late the availability time will not have too much meaning further on.

nigelmegitt:
Thanks @kozmaz87 that's a bit clearer in my mind. For semantic validation there will certainly be a need at some point to call a function on the `sequence` that might be called `recalculate_resolved_times` that traverses the sequence and applies the spec logic. I'm also assuming that each `EBUTT3Document` would probably cache its `earliest_computed_begin_time` and `latest_computed_end_time` to make this recalculation logic easier.

gilbrechbuhler:
Okay new proposal on the last commit : I just had this idea, not much to explain, the good thing here is that it allows for really few refactoring, with the default value of the parameter of `get_time` current calls like `clock.get_time()` are still valid and working as intended. And if you want the fix time, you just do `clock.get_time(True)`.

This is just a proposal, tell me what you think again. We can change the name of the new `current_time` method if you don't like it.

I deleted the test I added before because it is not useful to have a test on a proposal, I'll add one once a proposal has been validated.

nigelmegitt:
@gilbrechbuhler I don't really see the difference between this approach and the approach where you use a differently named function - in both cases the caller has to make a decision about which kind of clock value they want returned, rather than this decision being made elsewhere and abstracted away. That's why I'd prefer a flag on the clock to tell it what mode to use, so that can be set once somewhere and the rest of the code would not need to change. (I would also be happy to consider other designs using OO to achieve something similar)

gilbrechbuhler:
But there is something I don't understand here, in the end it is still the caller that will tell if it needs to use a real clock or a fake one. I mean, the flag has to be set by someone at a moment too, it is the node using the clock that will set it, no ? I don't understand how we could completely abstract that from the caller, as it is the caller that has the context to know if it needs a real time or the saved one.

gilbrechbuhler:
Actually my true question here : concretely, by who/where would this flag you talk about be set ? I think we do not picture it the same way at all in our respective minds haha. For me it would be something like the processing node that would tell what kind of time it needs.

nigelmegitt:
Right, yes, so you've got a `ConsumerNode` of some sort and you write a script that plugs in a `CarriageImpl` to it, and what you want to happen is that every time the `CarriageImpl` delivers a new document some processing happens (`on_new_data()`) - if that processing needs to know a clock time it should not need to know how time is progressing, just the time related to the data it is processing. Arguably `on_new_data()` should take a `timedelta` as well as the data. If anything needs to know some simulated 'time now' then it would make a call on the reference clock, which would have been set up to behave in an appropriate mode, perhaps by the script that created it. Spreading a requirement to know the clock mode everywhere that needs to get the time seems like it could get painful.

gilbrechbuhler:
@nigelmegitt well actually `data` can mean anything so we can simply send an array like `[timedelta, document]`. But this is another topic, let's not diverge too much.

Ok mmmh I see what you want, I think, I just did not look at that problem in the same angle at all.

gilbrechbuhler:
@nigelmegitt So, I think I understand what you want here. Could you look at the last stage of the PR to tell me if I'm going the right way ?

Again all names are modifiable, we can improve this as needed, even use some OO principles to implement it in a better way if this way is not good enough.

gilbrechbuhler:
With the last commit we can trigger the flag where we want and the nodes/carriages still have to just use get_time() without worrying about where the time they get comes from

nigelmegitt:
I think that works in principle, yes. @kozmaz87 are you happy with this approach?

gilbrechbuhler:
Documentation and better naming in the last commit.

gilbrechbuhler:
I wait on validation from everyone ( @kozmaz87 @nigelmegitt @frans-ebu ), then I will add some unit tests and then it will be merge-able.

nigelmegitt:
OK I'm happy with this - I'm not actually sure that we need to _use_ it but it's fine as is. If we discover later that we can do everything we need without using it then we can factor it out again, but I suspect it will be very useful for testing.

kozmaz87:
ACK. It can go.

Issue #83, opened 2016-07-21
Milli/Microseconds are not parsed when converting from times in documents to timedelta


gilbrechbuhler:
Discovered while writing tests for #69.
File containing the code : [https://github.com/ebu/ebu-tt-live-toolkit/blob/master/ebu_tt_live/bindings/_ebuttdt.py]

```
_groups_regex = re.compile('([0-9][0-9]+):([0-5][0-9]):([0-5][0-9]|60)(?:\.[0-9]+)?')

    @classmethod
    def as_timedelta(cls, instance):
        """
        Using regex parse value and create timedelta
        :param instance:
        :return:
        """
        hours, minutes, seconds = cls._groups_regex.match(instance).groups()
        return timedelta(hours=hours, minutes=minutes, seconds=seconds)
```
As you can see we never use the milli/microseconds.

Issue #82, opened 2016-07-21
Conversion from time in documents to timedelta does not work


gilbrechbuhler:
Discovered while writing tests for #69 : let's take an example to explain the problem, so the conversion from `limitedClockTimingType` to timedelta is implemented this way for now :

```
_groups_regex = re.compile('([0-9][0-9]+):([0-5][0-9]):([0-5][0-9]|60)(?:\.[0-9]+)?')

    @classmethod
    def as_timedelta(cls, instance):
        hours, minutes, seconds = cls._groups_regex.match(instance).groups()
        return timedelta(hours=hours, minutes=minutes, seconds=seconds)
```
Now this error happens when calling for example `document._ebutt3_content.body.begin.timedelta()` (if the time in the begin attribute of the body uses `limitedClockTimingType`) :

```
         hours, minutes, seconds = cls._groups_regex.match(instance).groups()
>       return timedelta(hours=hours, minutes=minutes, seconds=seconds)
E       TypeError: unsupported type for timedelta seconds component: unicode
```
We need to convert the extracted values before creating the timedelta object.

Note that the same problem exists with `fullClockTimingType`.

[https://github.com/ebu/ebu-tt-live-toolkit/blob/master/ebu_tt_live/bindings/_ebuttdt.py]

gilbrechbuhler:
@kozmaz87 do you think I can try to fix this issue along with #83, or do we risk big conflicts with your validation branch at some point here ?

Issue #69, opened 2016-07-14
Create time metric tests (for smpte)


nigelmegitt:
See #67 and #68: issue is to create tests that verify that the correct time units are used and the code does not get `ms` and `m` confused with each other.

gilbrechbuhler:
I don't close this one finally because we miss the checks for smpte, but as conversion from smpte is not implemented yet at all, it will have to wait.

Waits on #70

nigelmegitt:
SMPTE time expressions are no longer required. Closing.

PR #86, opened 2016-07-25
Create time metric tests and major fix conversion from document times to timedeltas

gilbrechbuhler:
Major fix for converting times from documents to timedeltas with tests. Fix for issues #69 , #82 , #83


kozmaz87:
This is good. ACK.

nigelmegitt:
OK, will merge later this morning unless anyone shouts at me not to!

PR #96, opened 2016-07-25
surfaced as part of the semantic validation development and since thi…

### LINKED ISSUE #95: Buggy test for timeBase attribute mandatory
SMPTE cannot have timecount timing type.

kozmaz87:
…s buggy test is on master here is a hotfix. This fixes #95


gilbrechbuhler:
Ouch I missed that, thanks for finding this bug. ACK, ok to be merged for me.

kozmaz87:
ok here here merge this one :)

nigelmegitt:
Looks good to me. It's intentional that the smpte times evaluate to different values than the non-smpte times, right?

kozmaz87:
of course. The bug being that timecount was used in smpte, which was deemed invalid by my semantic validation. The test does not deal with the actual values of the timings they just need to be valid to timebase that's all

nigelmegitt:
Merging this as @kozmaz87 needs it (and is nagging me) and also it's related to tests and appears harmless.

PR #103, opened 2016-07-26
Availability times in documents

### LINKED ISSUE #102: Add an availability time attribute to `EBUTT3Document` class
For #50

gilbrechbuhler:
fixes #102


nigelmegitt:
Looks good.

kozmaz87:
This is quite evil from coveralls. It marks the pull request with red cross like CI wouldn't have passed if the coverage decreases...

nigelmegitt:
Not evil - it's keeping us in order! We can still merge anyway, since we can't usefully test this right now.

Do we need to call the setter from the `carriageImpl`s as part of this PR and add tests that they execute properly, or do you want to pick that up elsewhere?

gilbrechbuhler:
Oh indeed I completely forgot to use it in the carriage implementations... I'll add that now. For tests I don't know, the lines I added are basically an attribute along with its getters and setters, I don't know if it needs testing at all. Maybe just checking that an exception is raised when something else than a timedelta is given.

nigelmegitt:
That would do it!

kozmaz87:
yes I believe that it would do. Assuming we don't want to feed it values of string but I guess at that layer of abstraction the canonical timedelta type is a better choice

gilbrechbuhler:
Yes I thought about it when writing the code. We could give a string and the document would then be responsible for converting it to the right timedelta value. But here I thought that it would be better to trust the clock giving the ability time, as this same clock will give us a timedelta respecting the chosen timebase.

kozmaz87:
So you are suggesting in the long run to take out the timedelta conversion from the bindings and put it in the document?

gilbrechbuhler:
No no, not what I meant, I speak only for the availability time here. This is why I think the document should get a timedelta for the availability time and not a string, otherwise we would have to convert the time in the `@availability_time.setter` method and that would be confusing. Because availability time never exists in the xml

kozmaz87:
I agree

gilbrechbuhler:
So I added the test and now the existing carrier of the simple consumer sets the availability time when it receives documents

gilbrechbuhler:
I am going to need this one for the filesystem carriage consumer part :)

kozmaz87:
ACK

nigelmegitt:
Needs conflicts to be resolved.

nigelmegitt:
All looking good. Merging.

nigelmegitt:
(ignoring -0.1% coverage drop)

Issue #85, opened 2016-07-22
Problems with timingtype matching when parsing XML


kozmaz87:
The following problem has arisen: Now that I put in the semantic checks for the timeBase->timingType things started to fail. In most cases the parser cannot decide whether it needs to instantiate a limited or a full clock time since the values look the same.

Issue #72, opened 2016-07-15
validation framework for semantic rules


kozmaz87:
We need a clean and concise way to do semantic validation. The aim so far seems to be that we would like to have this in a separate place from the main pyxb binding codebase.

LXML and XPATH has been discussed... @nigelmegitt  did not feel confident about it. me neither.

Current approach is to hook into pyxb's validation flow. It is possible to separate out the semantics and link them to the normative specification in [SPEC-CONFORMANCE.md](https://github.com/ebu/ebu-tt-live-toolkit/blob/master/testing/SPEC-CONFORMANCE.md)

PR #106, opened 2016-07-26
Kozmaz87/validation

kozmaz87:
This will affect:
#85 #72 

Once the pull request for the rest buggy test is merged this will pass


kozmaz87:
Ok guys. Have fun reviewing this. I am adding some extra tests and obviously some more documentation to it because this is dark magic here. In short the library(PyXB) did not offer out of the box the nice propagation of a parsing context so using python's threadLocal objects I monkeypatched the processing pipeline in the appropriate places to get it do what I wanted it to do for me. @gilbrechbuhler if you don't mind it would be very good for you to take a look and give me an impression how difficult it is to follow what's happening here. There could be an easier solution created if we forked PyXB and added the context object in the framework itself.

nigelmegitt:
Just looked, and for me anyway it does seem hard to follow initially. Yes please to more documentation!

gilbrechbuhler:
I think I understand most of it. However, after one month, I'm starting to get comfortable with pyxb and how we make all that work, so it makes this new part easier to understand. So I think some doc would be great indeed.

nigelmegitt:
Reminder: there are conflicts on this branch that need to be resolved before it can be merged.

nigelmegitt:
@kozmaz87 I think you're still working on this? I don't want to merge if you're going to make some more test or documentation commits.

nigelmegitt:
Merging this as "good enough" for now.

Issue #89, opened 2016-07-25
Calculate computed document activation times


EyalLavi:
Calculate computed activation begin and end times for the simple consumer. This is a subtask of #42.

kozmaz87:
I changed the wording to computed activation begin and end times because that is what the commits do to this ticket. The resolved activation times are in sequence context and will be another ticket.

PR #110, opened 2016-08-02
Kozmaz87/timing resolution

kozmaz87:
Let's start the review process on this as this is quite a complicated and to most of us very convoluted piece of this software. The current status is #89 is mostly resolved but another ticket will go into this that is the resolved activation times in the context of a sequence because #89 does only compute per document.


testing/bdd/features/timing/resolved_times.feature
```diff
@@ -0,0 +1,24 @@
+@timing @timingresolution
```
gilbrechbuhler: Are these tags just informative or do they have a real meaning for the tests ?

testing/bdd/features/timing/computed_times.feature
```diff
@@ -0,0 +1,48 @@
...
+    And it has p end time <p_end>
+    And it has span begin time <span_begin>
+    And it has span end time <span_end>
+    And the document is generated  # implicitly means it is valid.
+    And it has availability time <avail_time>
+    Then it has computed begin time <computed_begin>
+    And it has computed end time <computed_end>
+
+    Examples:
+    | time_base | body_begin  | body_end     | body_dur | div_begin   | div_end     | p_begin    | p_end      | span_begin  | span_end    | computed_begin | computed_end | avail_time  |
```
nigelmegitt: Seems like this is a good start, but it would be useful to have more than one p, or span, say, as children of the same parent, to check that the `par` semantics are correctly being handled.

gilbrechbuhler: Oh good point, will add that

nigelmegitt:
This is looking good. Is there any point in doing this multiple siblings test at other levels than span, or is it obvious that if it works in one place correctly then it will work in other places? For example, we could have examples with 1, 2 and 3 elements with different begin and end times at each level of div, p and span. I understand that would be verbose, and lots of cases, so don't want to push tests that are actually pointless.

gilbrechbuhler:
I'm not sure, I think it is enough to test on `<span>` because the code will go through the exact same steps with other elements as siblings. @kozmaz87 Do you agree ? What could be worth doing however is add a third span element to make sure it works with more than two sibling elements.

nigelmegitt:
Just had a quick peek in here - I'm getting double log messages for every document, which is a bit unexpected:

```shell
$ ebu-simple-consumer
[INFO] (2016-08-10 14:35:05,597) in ebu_simple_consumer[32] - This is a Simple Consumer example
[INFO] (2016-08-10 14:35:05,598) in twisted[154] - Starting factory <ebu_tt_live.twisted.websocket.BroadcastClientFactory object at 0x10b207190>
[INFO] (2016-08-10 14:35:07,656) in document_logger[88] - Syntactic validation successful
[INFO] (2016-08-10 14:35:07,658) in document_logger[128] - Semantic validation successful
[INFO] (2016-08-10 14:35:07,659) in document_logger[88] - Syntactic validation successful
[INFO] (2016-08-10 14:35:07,660) in document_logger[128] - Semantic validation successful
[INFO] (2016-08-10 14:35:07,660) in ebu_tt_live.node.consumer[26] - Creating document sequence from first document <ebu_tt_live.documents.ebutt3.EBUTT3Document object at 0x10b255cd0>
[INFO] (2016-08-10 14:35:07,662) in document_logger[88] - Syntactic validation successful
[INFO] (2016-08-10 14:35:07,663) in document_logger[128] - Semantic validation successful
[INFO] (2016-08-10 14:35:07,663) in document_logger[36] - Document TestSequence1__2 received. Calculated activation: [14:35:08.606000; 14:35:09.606000]
[INFO] (2016-08-10 14:35:09,627) in document_logger[88] - Syntactic validation successful
[INFO] (2016-08-10 14:35:09,628) in document_logger[128] - Semantic validation successful
[INFO] (2016-08-10 14:35:09,629) in document_logger[88] - Syntactic validation successful
[INFO] (2016-08-10 14:35:09,630) in document_logger[128] - Semantic validation successful
[INFO] (2016-08-10 14:35:09,630) in document_logger[36] - Document TestSequence1__3 received. Calculated activation: [14:35:10.608000; 14:35:11.608000]
```

etc.

Is this what we are expecting? It makes it look as though every document is being syntactically and semantically validated twice.

kozmaz87:
It is not double log messages. It is double validation. And it is expected as the code is structured that way

nigelmegitt:
Oh, right, not for now, but I'm interested to know why we do that.

ebu_tt_live/bindings/validation.py
```diff
@@ -159,43 +194,155 @@ def _post_timing_set_attribute(self, attr_use):
...
+        # If the forward running part of the traversal could not assign an end time we can use the backwards route
+        # which is in a way similar to dynamic programming because we take the children computed times and take the
+        # maximum value from them. SPECIAL case: a single leaf element with an undefined end time renders the entire
+        # branch undefined
+
+        if end_timedelta is not None and self.computed_end_time is None \
+                or end_timedelta is None and self.computed_end_time is not None:
+            # This is just a simple sanity check. It should never be triggered.
+            # Should the calculation be changed this filters out an obvious source of error.
+            raise LogicError()
```
gilbrechbuhler: I think I just understood why this line is making so much of our tests fail. When the document is first created, the validation is runned once. Then, in the tests, we set the availability time, which triggers the validation and the computing of times again. 

However, imagine I have a `div` without any timing information and a `p`inside it with an explicit `end`. The first time the document was validated, the `div` element actually got a `computed_end_time` in the end : the end time of its last (in the sensse of the timeline) child. So the second time we do the validation, when we set the availability time, when we get to validate the `div` element, it has no `end` attribute, hence `end_timedelta` is `None` however it has a `computed_end_time` that was defined during the first validation as explained before. And this is no an error.

So I think we should either change this `if` or find a way to solve that problem. I'm looking into it.

gilbrechbuhler: And so I wonder if the best solution is would not be to simply add :

```
self._computed_begin_time = None
self._computed_end_time = None
```
At the beginning of the `_semantic_preprocess_timing` method.
As everything is computed again when we run the validation the second time, it does not create problems, at least no problems I can see and it solves our `LogicError()` problem.

@kozmaz87 please tell me what you think about that

ebu_tt_live/documents/ebutt3.py
```diff
@@ -81,9 +178,100 @@ def availability_time(self, value):
...
+                        if min_child_begin is None:
+                            min_child_begin = current_child.computed_begin_time
+                        else:
+                            if current_child.computed_begin_time < min_child_begin:
+                                min_child_begin = current_child.computed_begin_time
+                    current_child_children = current_child.orderedContent()
+                    for child in current_child_children:
+                        bfs_queue.append(child.value)
+            if min_child_begin is not None:
+                self._computed_begin_time = min_child_begin
```
gilbrechbuhler: @kozmaz87 all this block is the way I handled computation of the computed begin times of the document. I handle it here because after the discussion with Nigel I realized that the document's computed begin time can be different from the body's active begin time. So now the document's computed begin time is the active begin time of the first element declaring a `begin` attribute if there is one such element.

kozmaz87: I have yet to reach that level of understanding. Can you point at a testcase?

ebu_tt_live/bindings/validation.py
kozmaz87: that is good. it is what I expected. good

ebu_tt_live/bindings/validation.py
```diff
@@ -159,43 +194,223 @@ def _post_timing_set_attribute(self, attr_use):
...
+                    # Try to assign it the last specified ancestor
+                    self._computed_end_time = self._semantic_dataset['timing_end_stack'][-1]
+            else:
+                children_computed_end_times = [item.computed_end_time for item in children]
+
+                if None in children_computed_end_times:
+                    # The endless document case propagates up
+                    self._computed_end_time = None
+                else:
+                    # Propagate the longest end time among the children
```
nigelmegitt: Might be worth adding a comment that this is where the `endsync="all"` semantic is coded.

ebu_tt_live/bindings/validation.py
```diff
@@ -159,43 +194,223 @@ def _post_timing_set_attribute(self, attr_use):
...
+    def _pre_init_variables(self, dataset, element_content):
+        super(BodyTimingValidationMixin, self)._pre_init_variables(dataset, element_content)
+        self._dur_timedelta = self.dur and self.dur.timedelta or None
+
+    def _post_cleanup_variables(self):
+        del self._dur_timedelta
+        super(BodyTimingValidationMixin, self)._post_cleanup_variables()
+
+    def _pre_calculate_end(self):
+        # This is all for the body element because of the dur attribute
```
nigelmegitt: If all those ifs and elifs are designed to test all the possible combinations of `dur` present and `end|begin` being absent or present the structure is maybe harder to read than it could be. Why not:

```python
        if self._dur_timedelta is not None
            # test all the combinations of begin and end being absent or present, setting proposed_end as we go
            ...
            # assign the value
            self._pre_assign_end(proposed_end)
        else
            # Fallback case if there is no duration specified the same as the other containers
            super(BodyTimingValidationMixin, self)._pre_calculate_end()
```

for example? Just wondering if there are secret cases or magic in here that are not obvious...

nigelmegitt:
This is a great piece of work - needs timing validation process and class model documentation and to address the two fairly minor comments I made, then good to merge IMO.

kozmaz87:
OMG looking at this now all those 27 files and over 2000 lines I am so glad that this feature is finally coming to a completion state...

nigelmegitt:
This is looking good. The documentation can be improved, but I'll raise a separate issue for that and merge this if everyone else is happy. @gilbrechbuhler @frans-ebu happy for me to move on with this Big Kahuna Merge?

nigelmegitt:
I'm going to merge this...

Issue #114, opened 2016-08-04
User Input Producer : send documents asynchronously 


gilbrechbuhler:
Send documents on a "looping task" basis, every three seconds for example and capture the changes made by the user "live".

For #20

Issue #113, opened 2016-08-04
User Input Producer : send documents on predefined times


gilbrechbuhler:
Have a running clock sending documents at times defined by the user.

For #20

Issue #112, opened 2016-08-04
User Input Producer : Add handling of times


gilbrechbuhler:
Add the possibility to set :
- [x] timeBase
- [x] clockMode
- [x] dur (at body level)
- [x] begin/end (at body level for a start)
- [ ] timings for regions when #117 is done
to created documents.

For #20

nigelmegitt:
I've split `dur` out because it's easier and can be static, and is probably more useful first than begin and end.

gilbrechbuhler:
First two points validated in commit b0d87431c23f1f92ec1633ec25ee6a86eb305659

gilbrechbuhler:
How much do we check user inputs on the UI side ? For the duration for example, I don't think it is a good idea to reimplement any logic concerning timing types here. I would just let the user enter whatever he wants, then show him an error when the document couldn't be validated on server side. What do you think ?

For now I just do one check : the user cannot send any document that does not have a sequence identifier.

nigelmegitt:
I think make it lightweight at first, later you'll probably end up implementing something a bit nicer anyway. So no problem right now to leave out timing type validation.

Issue #101, opened 2016-07-26
Develop a demo for the User Input Producer


gilbrechbuhler:
Task for #20 
Develop a demo to show the User Input Producer in action.

nigelmegitt:
We need to come back to this from a design perspective - I think for a generalised UIP we would need to do this but it might be complex. Another option is to use a single carriage mechanism and make the UIP send XML documents directly to it, using e.g. WebSocket. That could then be picked up even on the local machine. If we want to transfer to another carriage mechanism then we'd have a consumer and producer node that does:

```
on_new_data(self, data)
    node.process_document(data)

...

TransferNode.process_document(self, data)
    outputImpl.emit_document(data)
```

In other nodes the "TransferNode" has two CarriageImpls, one for inputs and the other for outputs, which can be independent of each other.

Then in the UIP we can just have js that opens a WebSocket, creates new documents based on user input and sends them to the WebSocket; the TransferNode sends those documents on.

This is my thinking at the moment, to make the UIP as simple as possible and still have document level validation present.

nigelmegitt:
Reviewing this issue, I think we now have enough to close this. We can currently:

1. Run `ebu-user-input-forwarder`
2. Open the `user_input_producer.html` page
3. Open in a separate window the `test.html` page
4. From the UIP page create a sequence and send documents
5. View those pages being validated in the user input producer process output
6. View the documents arriving on the test.html page

Plus we can even do this with 1x user input forwarder and _n_x sequences, generated by _n_ user input producer pages and viewed by _m_ test html pages. 

And, if we want we can show transition from one carriage mech to another by setting the user input forwarder to output to file system and then by running `ebu-simple-consumer` to show the documents' arrival being observed in the manifest file and the documents being processed.

We can also show documents being generated on demand, at scheduled times or asynchronously.

To me, this constitutes more than enough to demo the UIP!

Any other views before I close this issue?

nigelmegitt:
One thing that would be good to do before closing this is to update the Demo bit of the documentation to include the above demo possibilities.

gilbrechbuhler:
Will do

PR #120, opened 2016-08-08
User input producer up to #114

### LINKED ISSUE #97: Define precisely the functionalities expected from the user input producer
First step of #20 : I think before starting developing it the component we need to define precisely what is expected from the User Input producer component.

gilbrechbuhler:
Not finished yet but I do a pull request so everyone can review what has already been done.

Fix #97 #98 #101 #112 #113 #114


gilbrechbuhler:
So last push is quite big. Some changes happened. The time base and all related options are now to be set when creating a new sequence. I did that because otherwise it was possible to create wrong sequences (documents with different time bases in the same sequences for example).

You can now choose between sending documents live when clicking the Send button, or schedule document sending. When the schedule is set on "Local clock" it uses the computer's clock, when it is set on "media" it uses the local clock with an offset. So when in media mode, you have to click on "Synchronize" button before sending documents to synchronize the "00:00:00" of the media clock to the current time of the computer. Then, still in media mode, if you set "00:00:10" as a scheduled time, it will send them ten seconds after the synchronization.

You can also send documents asynchronously now, you can set the number of seconds between each send.

Please review and comment, I know the html file is big, but actually most of the javascript is there for the interface : showing fields x and y when option z is selected, ... .

It certainly contains bugs, I tested some scenarios and did not discover bugs in this version, yet.

gilbrechbuhler:
Now you can see the media clock running live once the synchronization is done.

frans-ebu:
@gilbrechbuhler Many tnx for this piece of work! I am happy to see this merged.

One minor comment: I think it would be useful to add the URL of the input-producer to the readme, so a naive user can start to use it immediately.

gilbrechbuhler:
@frans-ebu Done, good remark. I added this line to the README (in the Scripts part) : " To run it, just run `ebu-user-input-consumer` and open the file `ebu_tt_live/ui/user_input_producer/user_input_producer.html file.`"

gilbrechbuhler:
I updated the documentation a bit also, to explain the new ways of send documents (scheduled time, asynchronous)

README.md
```diff
@@ -32,7 +32,9 @@ and assumes the environment variables that the activation script sets):
 
     source ./env/bin/activate
 
-After having created the python virtual environment and having activated it the package
+To build the project you will also need node.js. Please read the instructions for your system [here](https://nodejs.org/en/download/package-manager/).
```
nigelmegitt: Is this accurate? I don't have node.js installed but I do have npm - they seem somewhat independent?

nigelmegitt:
Review comments from meeting:
@kozmaz87 : uip is good to go
Me: this is good to merge, with the one minor query about node.js vs npm in the documentation.

Plus @EyalLavi has raised #122.

README.md
gilbrechbuhler: For me npm was installed as a part of Node.js. Do you know a way to install npm only ? Because I did not find if there is.

gilbrechbuhler: Forget previous remark, we need Node.js. We use it to have npm but the scripts that precompiles the templates needs Node.js on the system anyway.

nigelmegitt: ok I must be misremembering - I thought I might have done something like `brew install npm` but when I do `brew list` it tells me I have node but not npm. Sorry about that.

nigelmegitt:
Also #124, which is now fixed too.

PR #132, opened 2016-08-16
User input forwarder

### LINKED ISSUE #126: Create a Distributor node
The User Input Producer (#20) acts as a WebSocket server and just consumes the incoming documents from the UIP HTML page. So that the documents generated can be used further downstream, e.g. encoding them, viewing them on the `test.html` page etc, it would be very useful to have a "forwarder", i.e. a Distributing Node from tech3370, which would act as a server in the same way for allowing incoming connections to receive documents but also allow incoming connections to send documents (if a WebSocket based CarriageImpl is used), or save them to the file system if a filesystem based CarriageImpl is used.

gilbrechbuhler:
A first try at the user input forwarder. So we have a new node that receives the document and forwards it further (by calling `emit_document` in its `process_document` function).

I created a new carriage class inheriting from `CombinedCarriageImpl` and having two attributes : a consumer carriage and a producer carriage. This allows me reuse existing code by using existing carriage implementations.

You can test it, start `ebu-user-input-forwarder` script, then open the User input producer UI and the `test.html` file. Create a sequence in the UIP and use it as the channel name in `test.html` when subscribing. Normally when you send documents from the UIP it shows up in `test.html`

@nigelmegitt : Is this the way you imagined this component ?
@kozmaz87 : what do you think of my solution ?

This version shows that changing the destination of the documents is really easy, simply modify the `ebu-user-input-forwarder` with another `sub_prod_impl` to have a new carriage.

@nigelmegitt : did you imagine a script with command line options to define how to handle documents ? (websocket, filesystem, ....). If yes, I can do it, it is quite easy to do.

Fix #126

### DIFF
--- a/ebu_tt_live/carriage/filesystem.py
+++ b/ebu_tt_live/carriage/filesystem.py
@@ -63,7 +63,6 @@ class FilesystemProducerImpl(ProducerCarriageImpl):
 
     _manifest_path = None
     _dirpath = None
-    _manifest_file = None
     _manifest_content = None
     _manifest_time_format = None
 
@@ -93,6 +92,9 @@ class FilesystemProducerImpl(ProducerCarriageImpl):
                 break
 
     def emit_document(self, document):
+        if self._manifest_path is None:
+            manifest_filename = "manifest_" + document.sequence_identifier + ".txt"
+            self._manifest_path = os.path.join(self._dirpath, manifest_filename)
         # Handle there the switch and checks to handle the string format to use
         # for times in the manifest file depending on your time base.
         filename = '{}_{}.xml'.format(document.sequence_identifier, document.sequence_number)
--- a/ebu_tt_live/carriage/forwarder_carriage.py
+++ b/ebu_tt_live/carriage/forwarder_carriage.py
@@ -1,7 +1,7 @@
 from .base import CombinedCarriageImpl
 
 
-class ForwarderCarriage(CombinedCarriageImpl):
+class ForwarderCarriageImpl(CombinedCarriageImpl):
 
     _consumer_carriage = None
     _producer_carriage = None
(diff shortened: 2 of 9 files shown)


nigelmegitt:
Yes, this is how I imagined the component. Nice! Command line options for configuring the in and out carriage implementations would be a good addition I think, though they would make the script dependent on all the possible carriage mechanisms. That's okay for now.

gilbrechbuhler:
`ebu--user-input-forwarder --folder-export <folder_name>` works now : writes the documents sent by the UI on the hard drive along with a manifest file.

nigelmegitt:
Just tried this out, and it works really well to forward to test.html, and demonstrate the validation happening (if you put an illegal time expression in, say, so the documents don't go through to the test.html page because the forwarder blocks them).

nigelmegitt:
And it works really well with the file system too! Overall, a BIG thumbs up from me.

nigelmegitt:
For reference, I ran these command lines with my sequence ID "floop":

Process 1:

```shell
ebu-user-input-forwarder --folder-export test_uip_forward
```

Process 2:

```shell
ebu-simple-consumer -m test_uip_forward/manifest_floop.txt -f
```

Every time I sent a document from the UIP page it appeared in Process 1 and then Process 2, in the terminal output. This is exactly what I was hoping for.

nigelmegitt:
I'm not sure if there's any test we can add here?

nigelmegitt:
Plan to merge this in the morning if there are no contrary views, probably just after the standup.

gilbrechbuhler:
I can indeed add some unit tests

gilbrechbuhler:
Unit tests are done

nigelmegitt:
Is there actually something that runs the test?

gilbrechbuhler:
They are unit tests so as long as their filename starts with `test`, the test cases start with `Test` and the test methods start with `test_`, they will be automatically found and run by py.test. `make test` runs these new tests already.

gilbrechbuhler:
Oups misclick sorry reopening

nigelmegitt:
Sorry, I forgot that. All good. About to merge.

PR #133, opened 2016-08-16
Refactor tests

### LINKED ISSUE #131: Refactor tests to make better use of step types
We use "given" steps everywhere in bdd tests, however we should use "given" to give you an object and then "when" when we modify a given object.

gilbrechbuhler:
Add some tags on tests and use `when` wherever it is more appropriate than `given`.

Fix #131


nigelmegitt:
Looks good to me

frans-ebu:
+1

PR #135, opened 2016-08-16
[Fix] Restrict UIP async sending interval value + bug fix

### LINKED ISSUE #134: User Input Producer : empty "asynchronous send time" field results in documents being sent as fast as possible
Add a check returning an error when the field is empty

gilbrechbuhler:
Fix #134 and fix a little bug not allowing to enter a fractional time (like 0.5). Add default value to asynchronous sending interval field.


nigelmegitt:
Looks good to me. This is minor, and non-complex, so I'm going to go ahead and merge immediately.

PR #143, opened 2016-08-17
Added requirements derived from 3350 v1.1

EyalLavi:


nigelmegitt:
I've made a slight tweak to fix R90 but this is looking good to me. Not sure how many of these requirements are already covered - I expect a lot of them are already in the XSD/pyxb syntactic validation, but they may not have explicit tests.

PR #150, opened 2016-08-18
Removed positive number restriction

EyalLavi:
(which belongs in 3380)


nigelmegitt:
Made a slight wording tweak to clarify that the data type is xs:decimal, other than that this is good.  Will merge.

PR #154, opened 2016-08-19
tts:padding tests

### LINKED ISSUE #152: Test validation for padding attribute

gilbrechbuhler:
Fix #152


ebu_tt_live/xsd1.1/ebutt_datatypes.xsd
```diff
@@ -65,7 +65,7 @@ Please note that the EBU-TT XML Schema is a helping document and NOT normative b
 	<xs:simpleType name="paddingType">
 		<xs:restriction base="xs:token">
 			<xs:pattern
-				value="(\d*\.?\d+(px|c|%))|([+-]?\d*\.?\d+(px|c|%))\s([+-]?\d*\.?\d+(px|c|%))\s([+-]?\d*\.?\d+(px|c|%))\s([+-]?\d*\.?\d+(px|c|%))|([+-]?\d*\.?\d+(px|c|%))\s([+-]?\d*\.?\d+(px|c|%))\s([+-]?\d*\.?\d+(px|c|%))|([+-]?\d*\.?\d+(px|c|%))\s([+-]?\d*\.?\d+(px|c|%))"
+				value="((\d*(\.\d+)?(px|c|%))|([+-]?\d*(\.\d+)?(px|c|%)))(\s((\d*(\.\d+)?(px|c|%))|([+-]?\d*(\.\d+)?(px|c|%))))*(\s((\d*(\.\d+)?(px|c|%))|([+-]?\d*(\.\d+)?(px|c|%))))*(\s((\d*(\.\d+)?(px|c|%))|([+-]?\d*(\.\d+)?(px|c|%))))*"
```
nigelmegitt: Aren't there a whole bunch of `[+-]?` bits missing from this? E.g. in the 2 value version the second one doesn't seem to be allowed a sign.

gilbrechbuhler: No for each of the for values you have to possibilities, separated by "|", one with the [+-], one without. The four values are separated by the "\s".

nigelmegitt: oh sorry I see what you mean. But then since the `[+-]` has a `?` after it, surely the two cases one with it and one without are unnecessary duplicates?

gilbrechbuhler: Oh you are right indeed... I always have troubles with regular expressions...

testing/bdd/features/validation/padding_data_type.feature
```diff
@@ -0,0 +1,70 @@
+@validation @syntax
+Feature: Padding Element and Datatype testing
+  tts:padding is only allowed on certain elements. values are constrained to one, two, three or four non-negative numbers appended by percentage “%”, c” or “px”, delimited by a space.
+
+  ## Assumes that we can pass <tag> to the examples so that we test that the attribute is only applied to the elements that support it.
```
nigelmegitt: Is this assumption valid? I mean, does this feature file generate all the permutations of scenarios for padding on both style and region?

testing/bdd/templates/padding_data_type.xml
```diff
@@ -0,0 +1,27 @@
...
+    </tt:metadata>
+    <tt:styling>
+{% if test_padding_syntax %}
+    <tt:style xml:id="ID001" tts:padding="{{value1}} {{value2}} {{value3}} {{value4}}"/>
+{% else %}
+    <tt:style xml:id="ID001" {{style_padding}}/>
+{% endif %}
+    </tt:styling>
+    <tt:layout>
+      <tt:region xml:id="ID002" tts:origin="0% 0%" tts:extent="100% 100%" {{region_padding}}/>
```
nigelmegitt: Why doesn't region also need the `if test_padding_syntax` check like style?

testing/bdd/features/validation/padding_data_type.feature
gilbrechbuhler: This comment was written by @EyalLavi . This part is just supposed to test that the document validation fails if you put the tts:padding attribute somewhere else than a region or style element and succeeds otherwise.

testing/bdd/templates/padding_data_type.xml
gilbrechbuhler: Because I test the permutations of the last 2 scenarios only on the style attribute, I can test them on the region attribute too.

nigelmegitt:
Oh I see, thanks, I had misunderstood. Now you can test for presence of padding attribute on correct/incorrect elements and you can test separately for different values on style and region (simultaneously). OK, thanks.

nigelmegitt:
Looks good to me.

PR #155, opened 2016-08-19
Bdd/smpte datatype

EyalLavi:


EyalLavi:
As you prefer. I realised this but thought it would be better to keep it separate for the workflow/blame trail.

kozmaz87:
You know you can blame by line in git right?

gilbrechbuhler:
The only problem by doing a new features file is that it needs to be loaded in the test file to be executed and it may be a bit confusing to test clock and media in the same file but not smpte. Other opinions ?

kozmaz87:
Afaik the test file can load multiple features right?

gilbrechbuhler:
The template needs to be updated to be used with smpte apparently, I'm on it. Tests were not running as the new feature was not loaded anywhere.

EyalLavi:
I can merge, it's easy enough. I'll work on the existing feature file now for R55 R56 R58 R59

kozmaz87:
@EyalLavi awesome stuff that you link the feature files back in the conformance document. Good idea

EyalLavi:
Thanks but the credit goes to Gil!

nigelmegitt:
So does this PR need more work or is it ready for review?

EyalLavi:
Based on Gil's comments, I'm moving all timebase format tests to the original file, so both PR155 and PR156 will be redundant.

nigelmegitt:
Closing PR155 and PR156, leaving the branches in place.

PR #158, opened 2016-08-19
timebase timing formats for smpte and missing tests on div

### LINKED ISSUE #157: Update timeformat tests with smpte and missing div tests

gilbrechbuhler:
Fix #157


testing/bdd/templates/timeBase_timeformat.xml
```diff
@@ -47,7 +50,22 @@
   {% endif %}
 {% endif %}
     >
-    <tt:div>
+    <tt:div
+{% if div_begin %}
+  {% if div_begin == "*?Empty?*" %}
```
nigelmegitt: Magic keyword? Think I've seen this before somewhere. Is this Gil's edit to allow pass-through of an empty value? Or does the `*?Empty?*` text need to appear actually within the scenario example? (it isn't in any of them so far)

nigelmegitt:
Why remove the smpte tests? Doesn't look like they've been moved anywhere else...

nigelmegitt:
Looks good to me. Will merge after merging the validation timing PR.

PR #182, opened 2016-08-30
Test R8 and R37

### LINKED ISSUE #185: Test R8 R37
Each document in a sequence shall have a distinct sequence number.

gilbrechbuhler:
Each document in a sequence shall have a distinct sequence number. Those tests test that this is ensured.

Fix #185


ebu_tt_live/documents/ebutt3.py
```diff
@@ -377,6 +378,12 @@ def _insert_or_discard(self, document):
         ):
             # This loop goes backwards and checks for trimmed documents
 
+            # We cannot have two documents with the same sequence number
+            if item.document.sequence_number == document.sequence_number:
+                raise SequenceNumberAlreadyUsedError(
+                    ERR_SEQUENCE_NUMBER_ALREADY_USED
+                )
+
             # If any found event is higher sequence number we quit
```
nigelmegitt: Are there scenarios where a duplicate (but old) sequence number is used and is not identified, because the loop exits before we reach it?

testing/bdd/features/validation/sequence_id_num.feature
```diff
@@ -1,6 +1,11 @@
 @validation @syntax @sequence @xsd
 Feature: Sequence ID and Sequence Number
-  Both are mandatory parameters and the sequence number has to be a number (no letters)
+  Both are mandatory parameters and the sequence number has to be a number (no letters).
+  Moreover two documents with the same sequence identifier shall have different sequence numbers.
```
nigelmegitt: The requirement is that two _distinct_ documents shall have different sequence numbers. Somewhere we should probably check that if a document arrives with the same sequence number and it is _identical_ to the previously received document with the same sequence number then it is discarded.

nigelmegitt: I just added https://git.ebu.io/ebutt/ebutt-part-3/issues/121 for this point, since the spec is potentially unclear.

testing/bdd/features/validation/sequence_id_num.feature
```diff
@@ -25,8 +30,39 @@ Feature: Sequence ID and Sequence Number
...
+    | testSeq1 | 999999999 |
+
+
+  # SPEC-CONFORMANCE: R8 R37
+  Scenario: Invalid sequence number for documents in the same sequence
+    Given a test sequence
+    And an xml file <xml_file>
+    When it has sequence number <doc1_seqnum>
+    And doc1 is added to the sequence
+    And we create a new document
```
nigelmegitt: a new _different_ document ?

testing/bdd/features/validation/sequence_id_num.feature
gilbrechbuhler: Oh so indeed I missed that point, I can compare the produced xml of both document, and raise the error only if they are different.

testing/bdd/features/validation/sequence_id_num.feature
gilbrechbuhler: I will correct that, I missed the *different* part

ebu_tt_live/documents/ebutt3.py
gilbrechbuhler: I will check that while doing the other corrections

gilbrechbuhler: The problem of comparing documents raised in #202 might appear here too, I planned on comparing the xml here, so we need to solve the problem in #202 first.

testing/bdd/features/validation/sequence_id_num.feature
nigelmegitt: This is a situation where we can usefully influence the spec - what is a useful and easily implementable way to define equality? @kozmaz87 @tairt

gilbrechbuhler: So sequence identifier and sequence number are the most obvious attributes to compare documents, but is it enough ? Because if it is explicitly authorized to send the exact same document twice (it might happen actually, in case of carriage defect for example,) then we need to also use other attributes to compare documents. @kozmaz87 @tairt @nigelmegitt 

For example the document would have to use the exact same timings and to have the exact same content.

nigelmegitt: In this case, it's clearly not enough to check sequence id and number - the equality requirement is not scoped, so it should apply to everything. Especially here where the requirement is that when the same sequence id and number are used the documents shall not be different.

gilbrechbuhler: The only way I see to apply the comparison to everything is to compare the xml produced by the bindings directly. @kozmaz87 do you see any other way ?

gilbrechbuhler:
Okay so before merging this one we should validate the solution used in #202 and merge it

gilbrechbuhler:
Corrections are ready, they will be pushed as soon as #202 has been merged if we agree on the solution found for #202.

ebu_tt_live/documents/ebutt3.py
```diff
@@ -415,6 +422,13 @@ def _insert_or_discard(self, document):
 
         for item in self._timeline.irange(this_begins):
             # This loop goes forward looking at offending events
+
+            # We cannot have two documents with the same sequence number
+            if item.document.sequence_number == document.sequence_number:
```
gilbrechbuhler: Line to be updated once #202 has been merged

gilbrechbuhler:
Steps to make this one mergeable :
- Wait for #202 to be merged
- Merge master here
- Modify the commented line (427 of `ebu_tt_live/documents/ebutt3.py) to add hash comparison

nigelmegitt:
See note on the original issue #182 - it looks like there's a mismatch between the requirement number and what this issue is trying to solve. Not sure why.

nigelmegitt:
This will need some more work once #182 is changed to match the spec.

EyalLavi:
Closing this PR since the requirements have been deprecated. See #297

PR #184, opened 2016-08-30
Bdd/sequence identical timing model

### LINKED ISSUE #183: Test R11

gilbrechbuhler:
Test for R11 : Every document in a sequence shall have an identical timing model as defined by using the same values for the `ttp:timeBase` and `ttp:clockMode` attributes. (Note issue to add `frameRate`, `frameRateMultiplier` and `dropMode` to this attribute set)

Fix #183

### DIFF
--- a/ebu_tt_live/clocks/local.py
+++ b/ebu_tt_live/clocks/local.py
@@ -5,6 +5,7 @@ from .base import Clock
 class LocalMachineClock(Clock):
 
     _time_base = 'clock'
+    _clock_mode = 'local'
 
     def get_real_clock_time(self):
         now = datetime.now().time()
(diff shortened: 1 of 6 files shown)


gilbrechbuhler:
I will actually close this one for now. I skip the invalid tests because the code does not check anything else than the sequence identifier for now, but adding timebase and related attributes checks is not hard, I will do that and reopen it. Maybe only SMPTE will be a problem here, at worst I just skip SMPTE tests.

gilbrechbuhler:
it is the best I can do for now, SMPTE is not truly implemented. I skip all the tests except one because we only have media and clockMode=local clocks implemented. There is some work to do to have everything needed for those tests. I propose to add them and unskip them in the future when utc and gps clocks will have been implemented.

nigelmegitt:
Not sure of the status of this - is more work going to happen on it now that we have utc clock?

gilbrechbuhler:
Having utc clock would allow me to unskip some tests.

nigelmegitt:
Looks good to me.

PR #188, opened 2016-08-30
Bdd/timebase related elements

### LINKED ISSUE #187: Add test ensuring clockMode is mandatory when timeBase=clock

gilbrechbuhler:
The PR itself adds only a test ensuring that clockMode is mandatory when timeBase=clock. However it completes the spec-conformance document from R60 to R73b because R69 to R73a were already tested.

Fix #187


testing/bdd/features/validation/timeBase_clock_clockMode_mandatory.feature
```diff
@@ -0,0 +1,32 @@
...
+  # SPEC-CONFORMANCE: R73b
+  Scenario: Invalid ttp:clockMode
+    Given an xml file <xml_file>
+    When it has ttp:timeBase attribute <time_base>
+    And it has ttp:clockMode attribute <clock_mode>
+    Then document is invalid
+
+    Examples:
+    | time_base | clock_mode    |
+    | clock     |               |
```
nigelmegitt: Is it worth also adding a test for an empty `clock-mode`, since the code is able to check for that?

testing/bdd/templates/timeBase_clock_clockMode_mandatory.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" ?>
+<tt:tt
+        ebuttp:sequenceIdentifier="testSequence001"
+        ebuttp:sequenceNumber="1"
+        ttp:timeBase="{{ time_base }}"
+{% if clock_mode %}
+  {% if clock_mode == "*?Empty?*" %}
+        ttp:clockMode=""
```
nigelmegitt: Previous comment about empty `clockMode` related to this line. On second thoughts, should this line deliberately omit adding the `clockMode` attribute, rather than adding an empty one?

gilbrechbuhler: When there is nothing in the `Examples` table, the `{% if clock_mode %}` is false and if you look at the template, then ttp:clockMode is never here. If we put `*?Empty?*` in the examples tables, then it is there and is `ttp:clockMode=""`. I can add an `*?Empty?*` test in the example table, then both cases will be covered.

gilbrechbuhler: Done

nigelmegitt: Thanks, apologies I always get confused by the empty/missing thing first time I read these. Not sure why!

nigelmegitt:
Looks good to me.

PR #190, opened 2016-08-30
Bdd/3350 value types

### LINKED ISSUE #189: Test 3350 values types
Test R93 to R100

gilbrechbuhler:
As you can see I had to update some regular expressions in the XSD. For colors in RGB and RGBA, numbers were not restricted to the range [0, 255]. And the fontSize regex was missing some cases and allowing cases it should not have.

Fix #189


ebu_tt_live/xsd1.1/ebutt_datatypes.xsd
```diff
@@ -27,7 +27,7 @@ Please note that the EBU-TT XML Schema is a helping document and NOT normative b
 	<xs:simpleType name="fontSizeType">
 		<xs:restriction base="xs:token">
 			<xs:pattern
-				value="(\d*\.?\d+(px|c|%))|([+-]?\d*\.?\d+(px|c|%))\s([+-]?\d*\.?\d+(px|c|%))"
+				value="([+-]?\d*\.?\d+(px|c|%))(\s([+-]?\d*\.?\d+(px|c|%)))?"
```
nigelmegitt: From tech3350 v1.1 §4.5: 
> The content is constrained to one or two non-negative values of type ebuttdt:lengthType.

So the `[+-]` should be removed here.

testing/bdd/features/validation/3350-value-types.feature
```diff
@@ -0,0 +1,156 @@
...
+    | 3350_value_types.xml | rgba(0,0)       |
+    | 3350_value_types.xml | rgba(0,256,0,0) |
+    | 3350_value_types.xml | #00MM           |
+    | 3350_value_types.xml | 000000          |
+
+
+  # SPEC-CONFORMANCE: R94
+  Scenario: Valid extent values
+    Given an xml file <xml_file>
+    When it has extent attribute with value <extent>
```
nigelmegitt: This is extent on a `region`, right? Just asking because `tts:extent` is allowed on the `tt` element but in that case _must_ have pixel values. From https://www.w3.org/TR/ttml1/#style-attribute-extent :
> If `tts:extent` is specified on the `tt` element, then the width and height must be expressed in terms of two `<length>` specifications, and these specifications must be expressed as non-percentage, definite lengths using pixel units.

testing/bdd/features/validation/3350-value-types.feature
```diff
@@ -0,0 +1,156 @@
...
+    Given an xml file <xml_file>
+    When it has tts:fontSize attribute with value <font_size>
+    Then document is valid
+
+    Examples:
+    | xml_file             | font_size |
+    | 3350_value_types.xml | 1%  2c    |
+    | 3350_value_types.xml | 1.5px     |
+    | 3350_value_types.xml | 1c 0c     |
+    | 3350_value_types.xml | -1%  2c   |
```
nigelmegitt: This is not valid since it has a negative number.

testing/bdd/features/validation/3350-value-types.feature
```diff
@@ -0,0 +1,156 @@
...
+  Scenario: Invalid line height values
+    Given an xml file <xml_file>
+    When it has lineHeight attribute with value <line_height>
+    Then document is invalid
+
+    Examples:
+    | xml_file             | line_height |
+    | 3350_value_types.xml | hello       |
+    | 3350_value_types.xml | 1em         |
+    | 3350_value_types.xml | 1c 2c       |
```
nigelmegitt: Worth adding a negative value in here just to check that the "no-negative value of type `ebuttdt:lengthType`" requirement is checked.

testing/bdd/features/validation/3350-value-types.feature
```diff
@@ -0,0 +1,156 @@
...
+  Scenario: Valid origin values
+    Given an xml file <xml_file>
+    When it has origin attribute with value <origin>
+    Then document is valid
+
+    Examples:
+    | xml_file             | origin  |
+    | 3350_value_types.xml | 10% 10% |
+    | 3350_value_types.xml | 1c 1c   |
+    | 3350_value_types.xml | 1px 1px |
```
nigelmegitt: Worth adding some negative values, which are actually valid here.

ebu_tt_live/xsd1.1/ebutt_datatypes.xsd
gilbrechbuhler: Okay so I may have spotted an incoherence between 3350 and 3370, in 3370 the fontSizeType is defined like this :
`<ebuttdt:fontSizeType>
: <ebuttdt:lengthType> <whiteSpace> <ebuttdt:lengthType>?`

With :
`<ebuttdt:lengthType>
: <scalar> | <percentage>

<scalar>
: <number> <units>

<number>
: <sign>? <non-negative-number>`

From that syntax definition, a negative number in fontSize is allowed, I will correct that

testing/bdd/features/validation/3350-value-types.feature
gilbrechbuhler: Yes it is on a region, I will give a clearer name to the step.

ebu_tt_live/xsd1.1/ebutt_datatypes.xsd
gilbrechbuhler: I checked, it is defined the same in 3350. And indeed there is one line saying that it is `non-negative` lengthType. Maybe it would be good to modify the enhance the syntax with a `positiveLengthType` or something like that, because if like me you miss the line saying `non-negative lengthType` and you look directly at the syntax, you cannot guess that you are wrong

testing/bdd/features/validation/3350-value-types.feature
gilbrechbuhler: Okay, same problem with the lengthType here, because the XSD was built following the syntax and the >=0 part was missed. I can make a regex specific for lineHeightType to remove the negative part.

testing/bdd/features/validation/3350-value-types.feature
gilbrechbuhler: done

testing/bdd/features/validation/3350-value-types.feature
nigelmegitt: I've added this point to https://git.ebu.io/ebutt/tech3350-and-3360-docs/issues/8 - I'm happy to add a `nonNegativeLengthType` for this purpose (which would permit an optional leading `+` sign.

gilbrechbuhler: I crafted a temporary solution in the XSD I will comment on the corresponding line

ebu_tt_live/xsd1.1/ebutt_datatypes.xsd
```diff
@@ -44,12 +44,17 @@ Please note that the EBU-TT XML Schema is a helping document and NOT normative b
...
+		<xs:union>
 			<xs:simpleType>
 				<xs:restriction base="xs:token">
 					<xs:enumeration value="normal"/>
 				</xs:restriction>
 			</xs:simpleType>
+		<xs:simpleType>
+			<xs:restriction base="xs:string">
+				<xs:pattern value="\d*\.?\d+(px|c|%)"/>
+			</xs:restriction>
```
gilbrechbuhler: My temporary solution is to not use length type and simply define the right regular expression. It is not the best solution, but adding `nonNegativeLengthType` now in the XSD, while it is not already in the spec would not be a good idea I think

nigelmegitt:
@gilbrechbuhler please could you resolve the conflicts on this?

gilbrechbuhler:
Done

nigelmegitt:
Looking good, thanks!

PR #202, opened 2016-08-30
test for R12 on distributing node

### LINKED ISSUE #201: Test R12 for distributing node
A passive node shall NOT modify input sequences and shall only emit sequences that are identical (including the sequence numbers) to the input sequence(s)

gilbrechbuhler:
Spec numbers like R12 have to be done on every node of the type they treat. I mean, here, we need one test for each different passive nodes we have.
I made the test so that extending it with new nodes will be easy.

Fix #201


testing/bdd/test_passive_node_shall_not_modify_documents.py
```diff
@@ -0,0 +1,49 @@
...
+    xml_file = template_file.render()
+    document = EBUTT3Document.create_from_xml(xml_file)
+    test_context['document'] = document
+    test_context['node'].process_document(document)
+
+
+@then('the emitted document is identical to the received one')
+def then_emitted_document_identical(test_context):
+    emitted_document = test_context['node']._carriage_impl.last_emitted_document
+    assert emitted_document.get_xml() == test_context['document'].get_xml()
```
nigelmegitt: What exactly does the `get_xml()` function return here? I'm worried that this won't check byte-wise identity between the input and the output documents, which is what I'm expecting by "identical" here. Alternatively maybe we need a better definition of identity to allow for validation processing and document re-generation. I think @kozmaz87 has a view on this also. This could turn into a spec issue, if the identity test is not well enough defined.

gilbrechbuhler: get_xml() returns the xml produced by pyxb from the `EBUTT3Document` object, so if both `get_xml()` return exactly the same string (containing the full xml representing the document) then documents are identical.

I cannot test `emitted_document == test_context['document']` because in the current code this would only compare sequence numbers and sequence identifiers. But we want to check that whole the document remains unchanged. Maybe we indeed need to define bettter ways to compare documents.

gilbrechbuhler:
Okay tell me what you think of the new solution

nigelmegitt:
It looks clever @gilbrechbuhler but I'm not familiar enough with things like the `zip()` function to know what it actually does...

gilbrechbuhler:
Okay so the `zip` function works like that : This function returns a list of tuples, where the i-th tuple contains the i-th element from each of the argument sequences or iterables.

So if you have a list `a=[1, 2, 3, 4]` and `b=[5, 6, 7, 8]`, then `zip(a, b)` will return the list `[(1, 5), (2, 6), (3, 7), (4, 8)]`

gilbrechbuhler:
Here it allows to compare corresponding child nodes between both DOMs easily

nigelmegitt:
I'm wondering if this isn't all far too complicated and we shouldn't just hash each incoming document (raw data) and if we see another document with the same seq id and seq num but with a different hash then issue an error, otherwise if they have the same hash then discard all but the first received one and log the discard.

That way we're actually testing for identity in the received data, which is the cleanest reading of the spec right now. I think there may be something to investigate later about non-significant changes to data that aren't considered to change the identity relationship, but that sounds very complicated to implement, and when you're just processing incoming documents you want a quick decision.

gilbrechbuhler:
Yeah my solution is quite costly in terms of resources I admit, I tried to implement it nevertheless as the case of receiving a document with the same sequence number twice should be really rare, at least I think.

But I agree that a solution based on hashing would be far better, so the question will then be, how do we compute the hash of a document ?

nigelmegitt:
I don't know, I must admit I imagined there would probably be off the shelf hashing code in Python somewhere that we could reuse.

gilbrechbuhler:
Okay so neither pyxb or minidom provide a way to hash an xml tree correctly. Looking on google with keywords `python xmldiff` gives some solutions but they all have the problem that they need at what point to go through the full tree, and I think this is not avoidable. The second solution given [there](http://stackoverflow.com/questions/3007330/xml-comparison-in-python) is actually the most elegant I could find. But again, I'm pretty sure that the library that converts the xml to a dictionnary is going recursively through the full tree.

nigelmegitt:
What about hashing the incoming text document before parsing it, and storing the hash?

gilbrechbuhler:
To use hash we could implement the `__hash__` function in our `EBUTT3Document` object, but again to be sure to define a hash that uniquely and fully represents a document we would have to go through the full tree.

gilbrechbuhler:
Oh maybe this is possible, but imagine we have :

`
<a attr1="test" attr2="test2">
    <b>hello</b>
</a>
`

And then you get :

`
<a attr2="test2" attr1="test">
    <b>Hello</b>
</a>
`
Both documents have the exact same semantic, they are identical except for the attribute order and should be considered equal. Comparing or hashing the received string will not work in that case.

nigelmegitt:
I would consider those to be distinct documents that are different for this purpose.

nigelmegitt:
It's not a semantic test.

gilbrechbuhler:
Oh alright, so if we consider those documents to be different then it can work

nigelmegitt:
I think this check only makes sense for code receiving documents, not sending.

I'm taking a strict view of 'identity' here because the spec is not clear about the identity test. If this gives us a problem later then we should revisit it. In any case I need to raise an issue on the spec for this ambiguity.

gilbrechbuhler:
I implemented a solution with hashes following our discussion, tell me what you think about it

nigelmegitt:
Looks good. Are the tests running on a "passive" node? I'm curious to know if we actually have a truly passive node that passes.

gilbrechbuhler:
Yes of course, the distributing node is a passive node. Actually there will be the need to extend the tests for each new passive node but it should be as easy as copying the `given a distributing node` step and modify it a bit for the new passive node.

nigelmegitt:
So the test already runs and passes? I suppose I'm concerned that we take in XML, validate it, then regenerate it before output, whereas we should a) validate and b) send on the original. That way we preserve any foreign namespace content that has no bindings, for example.

nigelmegitt:
For that purpose a good addition to the test would be to deliberately include made-up content in a foreign namespace.

gilbrechbuhler:
Nodes handle document objects, and the xml is converted into document objects in the carriage. The distributing node does that :

```
def process_document(self, document):
        log.info(document)
        log.info(" " + str(document.sequence_identifier) + "_" + str(document.sequence_number))
        log.info(document.get_xml())
        self._carriage_impl.emit_document(document)
```

If we don't want to evaluate the xml at all in passive nodes, we actually need to change some things in the architecture : where we go from xml to document object.

gilbrechbuhler:
More precisely, when the consuming part of a carriage implementation receives xml, it constructs an `EBUTT3Document` object with it and gives that to the node's `process_document` method

gilbrechbuhler:
So the tests you propose won't work I guess as the content from the foreign namespace will not be validated by the bindings

nigelmegitt:
That's interesting. The code that calls `process_document()` is in something like `on_new_data()` if I remember correctly. If that is creating the document and then passing it on then it may indeed be doing that in the wrong place. Would make sense to push conversion to a document down the chain so that process_document() can be called and then the carriage impl's emit_data.

Something like (forgive my pseudo-python):

```
impl
    on_new_data(data)
        node.on_new_data(data)

   emit_data(data)
       pass

node
    on_new_data(data)
        document = EBUTT3Document(data)
        self.on_new_document(document)

    emit_document(document)
        data = document.to_xml()
        emit_data(data)

    emit_data(data)
        impl.emit_data(data)

distributingNode (node)
    on_new_data(data)
        super.on_new_data(data)
        self.emit_data(data)
```

Would that cause problems?

gilbrechbuhler:
I don't think it would cause big problems, however it is quite refactoring as it means modifying all carriage mechanisms and all nodes we have to follow this model. What I mean is I am not sure it is a good idea to do that in this Pull Request. And maybe @kozmaz87 had a specific reason to create the document in the `on_new_data` method of the carriage and not in the node in the first place, was there a specific reason ?

Raising an issue for that matter is the best idea here I think

gilbrechbuhler:
To be able to merge this pull request some steps are needed :
- Solve issue #216 and once the fix is merged in master, merge it here
- The test may have to be updated to follow the new architecture depending on the changes made on the carriages and nodes

testing/bdd/test_passive_node_shall_not_modify_documents.py
```diff
@@ -31,8 +60,6 @@ def when_processes_document(distributing_node, test_context, template_file):
 
 @then('the emitted document is identical to the received one')
 def then_emitted_document_identical(distributing_node, test_context):
-    from ebu_tt_live.utils import compare_xml
     prod_carriage = distributing_node.producer_carriage
     assert prod_carriage.emit_data.call_count == 1
     assert compare_xml(test_context['document'], prod_carriage.emit_data.call_args[0][0])
```
nigelmegitt: Really minor point, but using `args[0][0]` is less readable and less robust than `args[1]['data']` as on line 51.

nigelmegitt:
Thanks for that change - I think we need a comment on R118 in spec-conformance.md too about the partial test for `deep-equals`, please.

EyalLavi:
R118 does not exists in this branch. Opened #347 to add a note on deep-equal.

nigelmegitt:
The comment on R118 is now in place, so this is good to merge.

PR #203, opened 2016-08-31
utc clock

gilbrechbuhler:
Doing #184 I realised making an UTC clock was really easy in python, so here it is.

### DIFF
--- a/ebu_tt_live/clocks/__init__.py
+++ b/ebu_tt_live/clocks/__init__.py
@@ -1,6 +1,7 @@
 
 from . import base
 from . import local
+from . import utc
 from . import media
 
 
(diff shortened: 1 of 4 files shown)


nigelmegitt:
Looks good, I'm confused by the coverage decrease though - does the test not run?

gilbrechbuhler:
It runs but only covers the instanciation, it does not cover the `get_real_clock_time` method. I don't really now if this one is testable as it would be like testing pythons `utcnow()` method. However I just thought about one test I can add, will do that right now.

nigelmegitt:
Looks good to me!

nigelmegitt:
Merging, should facilitate #184

PR #205, opened 2016-09-01
Kozmaz87/ebuttd xsd2

kozmaz87:
XSD, the converter script, default values for conversion.


nigelmegitt:
Please could you move the non-passing cases to the 'invalid' table so we don't lose them?

nigelmegitt:
I'm picking up the illegal font size move task.

ebu_tt_live/node/consumer.py
```diff
@@ -46,3 +45,23 @@ def reference_clock(self):
     @reference_clock.setter
     def reference_clock(self, value):
         self._reference_clock = value
+
+
+class EBUTTDConverterConsumer(SimpleConsumer):
```
nigelmegitt: This can be done later, but I would change the name to "Encoder" to be consistent with the spec.

ebu_tt_live/scripts/ebu_ebuttd_converter.py
```diff
@@ -0,0 +1,52 @@
...
+parser.add_argument('-i', '--interval', dest='interval', metavar='INTERVAL', type=float, default=2.0)
+
+
+def main():
+    args = parser.parse_args()
+    create_loggers()
+    log.info('This is a Simple Consumer example')
+
+
+    consumer_impl = TwistedConsumerImpl()
```
nigelmegitt: Can be done later, but we should add the file system impl back in here so we can do this repeatably with stored files.

ebu_tt_live/xsd/ebutt_d.xsd
```diff
@@ -0,0 +1,315 @@
...
+		<xs:attribute ref="ttp:timeBase" use="required"/>
+		<xs:attribute ref="ttp:cellResolution"/>
+		<xs:attribute ref="xml:lang" use="required">
+			<xs:annotation>
+				<xs:documentation>The language for which the EBU-TT-D document is prepared. (STL
+					mapping: Language Code (LC))</xs:documentation>
+			</xs:annotation>
+		</xs:attribute>
+	</xs:complexType>
+	<xs:element name="ttd" type="tt:d_tt_type"/>
```
nigelmegitt: Should add a comment in here about why the element is deliberately misnamed!

nigelmegitt:
Thanks!

nigelmegitt:
This is incomplete but good to merge now so we can build further changes like fixing the XSD on it. Will do so.

nigelmegitt:
@kozmaz87 shall I delete the branch or do you want to keep working on it?

nigelmegitt:
I've raised #206 and #207 to avoid forgetting the line comments I added but did not resolve before merging.

kozmaz87:
branch can be deleted

PR #212, opened 2016-09-01
Bdd/document metadata elements order

### LINKED ISSUE #211: Test R38 and R39

gilbrechbuhler:
Fix #211 

Ensures that the metadata elements introduced in 3370 have to be in the right order and after any other metadata element in documentMetadata.


nigelmegitt:
Ah, this is useful. Thanks @gilbrechbuhler

nigelmegitt:
This is fine as is, but it would also be useful to include some metadata from tech3350, which is allowed to precede (but not follow) the tech3370 metadata in `documentMetadata`.

nigelmegitt:
That can be a separate PR to test the tech3350 requirement I guess.

gilbrechbuhler:
I did actually with documentIdentifier, but it could be enhanced with more indeed

PR #213, opened 2016-09-01
Nigelm/xsd attribute qualification

### LINKED ISSUE #204: XSD has incorrect namespace for attributes
Some attributes, like `style` on the `tt:style` element, are being defined with a qualified namespace by default when they should be in no (i.e. global) namespace. The `attributeFormDefault="qualified"` line in the `xs:schema` element should not be there, and if qualified attribute names are needed they should be specified locally.

nigelmegitt:
Fix #204 by removing namespace qualification from global namespace attributes.


nigelmegitt:
Review requested please @kozmaz87 - should be quite simple, though there are a lot of files. Essentially, it's a removal of namespace from begin, end and dur attributes everywhere, including where a namespace had been used in python code - I replaced it with `None` as per the pyxb documentation.

nigelmegitt:
I just did one further check on this, that a supplied file with foreign namespace content in `tt/head/metadata` still processes correctly. In doing this I discovered that the file system carriage mech cannot handle smpte timebase yet (see #215) and that the pyxb bindings generator successfully warns about foreign namespace content but does not break. Very good! 

For me I'm happy to merge this, but I'd really appreciate a positive comment - @kozmaz87 I think you've looked at this?

kozmaz87:
Yes this can go ahead.

nigelmegitt:
Thanks, merging.

PR #218, opened 2016-09-02
body styles missing

kozmaz87:


nigelmegitt:
Looks good to me. Am conscious that we don't have any test content that stresses this, but I just manually modified an example to reference a style from the `body` element and it validated correctly.

PR #219, opened 2016-09-05
[NO TICKET] Ericsson generated test sequence

nigelmegitt:
A sequence of documents generated by Ericsson’s live production tool,
fixed up by me to make them valid.

Note that this change does not modify any code, just adds a test resource that can be processed using ebu-simple-consumer.


nigelmegitt:
Note that this change does not modify any code, just adds a test resource that can be processed using ebu-simple-consumer.

nigelmegitt:
Going to merge this directly since it can't break anything!

Issue #220, opened 2016-09-06
empty ebuttm:documentRevisionNumber passes syntactic validation


nigelmegitt:
A document with an empty `ebuttm:documentRevisionNumber` is passed by the validation code but not by Oxygen with an XSD validation. The type is `xs:nonNegativeInteger`.

kozmaz87:
Yes this issue is caused by the default value PyXB implementation details of the type permit. It is a PyXB bug and needs working around

kozmaz87:
Related pyxb issue
https://github.com/pabigot/pyxb/issues/71

PR #221, opened 2016-09-06
[NO TICKET] Further simple example set

nigelmegitt:
Includes colour styles

Based on updated codebase. Includes data that relates to #220 - it should be invalid but gets syntactically validated.


nigelmegitt:
Merging since this has no impact on code.

PR #227, opened 2016-09-26
Kozmaz87/segmentation

kozmaz87:
Segmentation is almost ready to be pulled in. Completely accurate style conversion should go into another pull request as it will prolong this one way too long. This one deals with segmentation of document sequences and merging. Most of its design flaws have been corrected.


ebu_tt_live/documents/test/1.xml
```diff
@@ -0,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?><tt:tt xmlns:tt="http://www.w3.org/ns/ttml" xmlns:bbctt="http://www.bbc.co.uk/ns/bbctt" xmlns:ebuttdt="urn:ebu:tt:datatypes" xmlns:ebuttm="urn:ebu:tt:metadata" xmlns:ebuttp="urn:ebu:tt:parameters" xmlns:ebutts="urn:ebu:tt:style" xmlns:ericssontt="http://www.ericcson.co.uk/ns/ericcsontt" xmlns:ttm="http://www.w3.org/ns/ttml#metadata" xmlns:ttp="http://www.w3.org/ns/ttml#parameter" xmlns:tts="http://www.w3.org/ns/ttml#styling" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" ebuttp:authorsGroupControlToken="2" ebuttp:authorsGroupIdentifier="prerna_b" ebuttp:referenceClockIdentifier="bst" ebuttp:sequenceIdentifier="localhost EbuTT3 TestSeq" ebuttp:sequenceNumber="647" ttp:cellResolution="40 24" ttp:clockMode="local" ttp:frameRate="25" ttp:frameRateMultiplier="1 1" ttp:timeBase="clock" xml:lang="en-GB">
```
kozmaz87: Fair enough...

bbc-daniel:
As you say @kozmaz87,  without spending a significant amount of time grokking this, any review is going to be superficial. So on the subject of the superficial, have you considered adding pep8 checks to your build pipeline. As it stands there are 2877 violations not including line-length:
`$ pep8 --max-line-length=1000 | wc -l
2877`
If this project is expecting contributors, I would enforce style checks sooner rather than later.

kozmaz87:
ok PEP8 is useless in our case because neither Twisted nor PyXB is following it. I am following it in my own code. But those libs aren't and in many ways the generated code is non pep8

kozmaz87:
Reason being: Twisted and PyXB both predate PEP8

bbc-daniel:
Fair enough. On the topic of generated code. Does that need to be committed?

kozmaz87:
No it does not and it is not. But hooking into pyxb and twisted involves me having to use their calling conventions. Not to mention that the XML uses camelCase so pretty much whenever I am in the bindings module I am writing stuff like: div.style.fontSize

bbc-daniel:
Fair enough. On my naive reading of the makefile, I thought everything under /bindings was generated. apart from these points, there's not much I can realistically comment on without investing significant time, other than to say that your bdd suite is coming along nicely 😀

kozmaz87:
Yeah we managed to tackle most challenges with it. Actually I kinda hacked around sometimes I have a template_dict fixture and a test_context fixture because we were supposed to use only one given and multiple when statements according to best practices

kozmaz87:
The interesting bits are in the pyxb_utils.py file. I defined there a generic framework to get recursive operations walk the tree the same way.

bbc-daniel:
Which best practices are those. I've never seen an example using multiple whens. It's always:
Given ...
And ...
When ...
Then ...
And ...

kozmaz87:
To be honest no idea. Gil has read it somewhere so we changed some tests to work that way....

kozmaz87:
Not quite sure anymore but he ran into some sort of limitations with the multiple given statements

bbc-daniel:
I'd be very interested to see the source. Struggling to imagine a need for multiple whens. My first instinct would be that at some point along the way, someone has got their Givens and Whens mixed up.

nigelmegitt:
I am not sure without checking if there are also additions to SPEC-CONFORMANCE.md but if so they should be made alongside the new BDD tests. Reference #235 and #236.

testing/bdd/features/segmentation/segmenting_sequence.feature
```diff
@@ -0,0 +1,29 @@
...
+    And body begins at <body1_begin>
+    And document added to the sequence
+    And we create a new document with <body_dur> <span1_begin> <span2_begin> <span1_end> <span2_end>
+    And body begins at <body2_begin>
+    And document added to the sequence
+    And we create a new document with <body_dur> <span1_begin> <span2_begin> <span1_end> <span2_end>
+    And body begins at <body3_begin>
+    And document added to the sequence
+    And the sequence is segmented from <range_from> to <range_to>
+    Then the fragment contains body from <frag_body_begin> to <frag_body_end>
```
nigelmegitt: Worth clarifying these are computed body begin and end times.

Issue #244, opened 2016-10-13
Test basic scripts that they work


kozmaz87:
consumer, encoders, producer

kozmaz87:
Some scripts still need testing but to what extent. I am having difficulties testing things that would otherwise need to be mocked out... such as long running consumers and producers but we need to work this one out and specify what scenarios we would like to do.

nigelmegitt:
From conversations with @kozmaz87 this is generally a difficult thing to automate, so bumping to R3. Also the configurator means that the range of possible runtime permutations is now huge, but also hugely reduces the risk of bad configuration or script errors, so I'm less concerned about this now.

nigelmegitt:
Again, this is not going to be done in R3, bumping.

PR #245, opened 2016-10-17
Kozmaz87/small fixes

### LINKED ISSUE #242: Add http proxy support
Which I did just move it to a branch and merge it.

### LINKED ISSUE #243: Segmentation discard functionality. Make sure it works.
It is leaking memory for some reason. Investigate

kozmaz87:
This pull request contains minor ports of some code that was written for the demo branch and also adds testing for memory leaks adds discard functionality to the consumer classes.
closes #242 
closes #243


ebu_tt_live/scripts/ebu_ebuttd_encoder.py
```diff
@@ -39,6 +39,8 @@
                     default='current', metavar='HH:MM:SS.mmm')
 parser.add_argument('-o', '--output-folder', dest='output_folder', default='./')
 parser.add_argument('-of', '--output-format', dest='output_format', default='xml')
+parser.add_argument('--proxy', dest='proxy', help='Proxy server', type=str, metavar='PROXY:PORT')
```
nigelmegitt: Would be useful to clarify if the protocol prefix is expected here or not, e.g. should it be `http://server_path:port` or just `server_path::port` ?

ebu_tt_live/scripts/ebu_simple_consumer.py
```diff
@@ -30,6 +30,7 @@
                     help='Works only with -m, if set the script will wait for new lines to be added to the file once the last line is reached. Exactly like tail -f does.',
                     action="store_true", default=False
                     )
+parser.add_argument('--proxy', dest='proxy', help='Proxy server', type=str, metavar='PROXY:PORT')
```
nigelmegitt: As above, useful to describe if protocol is needed, in the help string.

ebu_tt_live/scripts/test/test_scripts.py
```diff
@@ -0,0 +1,12 @@
+
+from subprocess import Popen, PIPE
+from unittest import TestCase
+from ebu_tt_live.scripts import ebu_dummy_encoder
+
+
+class TestDummyScript(TestCase):
+
+    def test_simple_run(self):
+        process = Popen('ebu-dummy-encoder', stderr=PIPE, stdout=PIPE)
```
nigelmegitt: How does this work if the system firewall needs confirmation that the process should accept incoming connections?

kozmaz87: This is exactly the reason it is only the dummy encoder that is in this file for the moment. Twisted supports mocking out networking via their twisted.trial package and make it work. I will come to do it when we get a bit more time. #244 stays open for this reason.

nigelmegitt: Ah right, interesting.

nigelmegitt:
Looks like a summary of the command line parameters could be added to `docs/source/overview.rst` but is not essential here, so I'm happy with this as is.

Issue #233, opened 2016-10-06
write tests for delay node


malikbeytrison:

EyalLavi:
Provisionally assigning this to R4 although we may decide that variable delay can be added in a later release.

nigelmegitt:
R4 is correct since the delay node component is an R4 objective, and we need to make sure it is tested.

nigelmegitt:
This is pending PR merge (#305).

nigelmegitt:
Looking at spec-conformance post merge of #305 there are several requirements not yet tested, specifically R109-R117, or at least the BDDs for those are not on the file, so this is not actually ready yet.

nigelmegitt:
As part of #351 (specifically in 11ba71f214ac7cba106ebe97179111f3a9d60b63 and 5feb3cb) I have moved the skipped buffer delay node sequence number identity scenario out of the retiming delay node feature file and changed the comment to reflect that this now needs to test for document identity on emission rather than sequence number identity, ideally using `deep-equals` but forgivably using a lesser test.

EyalLavi:
@kozmaz87 could you please review https://github.com/ebu/ebu-tt-live-toolkit/commit/a8305a6f5532a98cc892c0f77465722186d6df0a as we discussed yesterday?

PR #246, opened 2016-10-20
Malik/issue 231 fixed delay node

malikbeytrison:


ebu_tt_live/node/delay.py
```diff
@@ -0,0 +1,71 @@
...
+
+    _reference_clock = None
+    _sequence = None
+
+    def __init__(self, node_id, carriage_impl, reference_clock):
+        super(FixedDelayNode, self).__init__(node_id, carriage_impl)
+        self._reference_clock = reference_clock
+
+        random_int = random.randint(0, 9999999999)
+        new_identifier = 'sequence_{0}'.format(str(random_int))
```
nigelmegitt: Might be nice to allow a desired sequence ID to be specified in the constructor. Also the delay period.

ebu_tt_live/node/delay.py
```diff
@@ -0,0 +1,71 @@
...
+
+        # change the sequence identifier
+        document.sequence_identifier = self._sequence
+
+        # TODO: add an ebuttm:trace element to the document metadata
+
+        fixed_delay_int = 2
+
+        # document is explicitly timed: modify the document
+        if document.binding.body.begin or document.binding.body.end:
```
nigelmegitt: We need to look at this test - I am not convinced it is sufficient. I think we need to traverse the tree to see if any elements have `begin` or `end` attributes not just the `body`.

ebu_tt_live/node/delay.py
```diff
@@ -0,0 +1,71 @@
...
+
+            update_children_timing(document.binding, document.time_base, fixed_delay_int)
+            self._carriage_impl.emit_document(document)
+
+        # document is implicitly timed: pause a while, re-emit later
+        else:
+
+            time.sleep(fixed_delay_int)
```
nigelmegitt: I'm not sure how twisted etc handle this: what happens if other documents arrive (or want to arrive) during this sleep? I wonder if there's a way to schedule the emission of the document for a known time instead of sleeping here, and if that would be better.

ebu_tt_live/node/delay.py
```diff
@@ -0,0 +1,71 @@
...
+        for child in children:
+
+            if hasattr(child.value, 'begin') and child.value.begin != None:
+
+                if timebase == 'clock':
+                    delay = LimitedClockTimingType(timedelta(seconds=delay_int))
+                    child.value.begin = LimitedClockTimingType(child.value.begin.timedelta + delay.timedelta)
+                elif timebase == 'media':
+                    delay = FullClockTimingType(timedelta(seconds=delay_int))
+                    child.value.begin = FullClockTimingType(child.value.begin.timedelta + delay.timedelta)
```
nigelmegitt: I think we may need to permit timebase="smpte" with clockMode="continuous", in which case we would need an extra line here I guess. Please could you add a TODO?

ebu_tt_live/node/delay.py
```diff
@@ -0,0 +1,71 @@
...
+
+            if hasattr(child.value, 'end') and child.value.end != None:
+
+                if timebase == 'clock':
+                    delay = LimitedClockTimingType(timedelta(seconds=delay_int))
+                    child.value.end = LimitedClockTimingType(child.value.end.timedelta + delay.timedelta)
+                elif timebase == 'media':
+                    delay = FullClockTimingType(timedelta(seconds=delay_int))
+                    child.value.end = FullClockTimingType(child.value.end.timedelta + delay.timedelta)
```
nigelmegitt: Same here - a TODO smpte would be a useful flag to show us where we need to add code to support smpte discontinuous.

ebu_tt_live/node/delay.py
```diff
@@ -0,0 +1,71 @@
...
+
+            update_children_timing(child.value, timebase, delay_int)
```
nigelmegitt: You don't need to update the children timing if you updated the begin, because that moves the syncbase for the children.

ebu_tt_live/node/test/test_delay.py
```diff
@@ -0,0 +1,8 @@
+from unittest import TestCase
+from ebu_tt_live.node.delay import FixedDelayNode
+
+class TestFixedDelayNode(TestCase):
+
+	def test_process_document(self):
```
nigelmegitt: Ah, this would be why coveralls is saying there's a drop in coverage!

ebu_tt_live/node/delay.py
kozmaz87: Yes there is a way to schedule them by using reactor timeouts or loopingcalls

ebu_tt_live/node/test/test_delay.py
malikbeytrison: Yes, this is planned in issue #233

ebu_tt_live/node/delay.py
malikbeytrison: I'm not sure to understand. The way I was thinking of this is that update_children_timing is called a first time on the document.binding, and recursively called by all the children in order to traverse all the document tree. If I remove line 71, the function will only execute on the head and body elements.

ebu_tt_live/node/delay.py
malikbeytrison: Fine, I can create a function similar to update_children_timing (tree traversal process) to do this test.

ebu_tt_live/node/delay.py
nigelmegitt: I mean you need to call update_children_timing only if you have not modified the begin time; if you have modified the begin time then all the descendants just move along with it and do not need to be updated. This is because the begin time is used to define the syncbase for the child elements.

ebu_tt_live/node/delay.py
nigelmegitt: Sounds good. @kozmaz87 is this code already written somewhere?

malikbeytrison: I added this functionality in 43f5b8e

ebu_tt_live/node/delay.py
malikbeytrison: Done in 2763171

ebu_tt_live/node/delay.py
malikbeytrison: Done in 2763171

ebu_tt_live/node/delay.py
malikbeytrison: Done in c6923d7

testing/bdd/features/timing/delay.feature
```diff
@@ -68,13 +68,16 @@ Feature: Delay of a document sequence
     | 00:00:05   |              |           |          |          |          | 00:00:10 | 00:00:15           |                  | 00:00:15          |                 | 00:00:15        |               |            |          |          |
```
nigelmegitt: In general for all of these the behaviour that we need to test is that the correct _specified_ times end up in the output document. I don't mind keeping these calculations of _computed_ times to help with debugging etc but we need the BDD to test the output documents' specified times as well.

ebu_tt_live/node/delay.py
```diff
@@ -25,16 +23,31 @@ def process_document(self, document):
 
         # TODO: add an ebuttm:trace element to the document metadata
 
-        # document is explicitly timed: modify the document
-        if is_explicitly_timed(document.binding):
+        update_children_timing(document.binding, document.time_base, self._fixed_delay)
+        document.validate()
```
malikbeytrison: It is done in line 22, right?

ebu_tt_live/node/delay.py
```diff
@@ -25,16 +23,31 @@ def process_document(self, document):
...
+        self._fixed_delay = fixed_delay
+        self._document_sequence = document_sequence
+
+    def process_document(self, document):
 
-            update_children_timing(document.binding, document.time_base, self._fixed_delay)
-            document.validate()
-            self._carriage_impl.emit_document(document)
+        # change the sequence identifier
+        document.sequence_identifier = self._document_sequence
```
malikbeytrison: Done in 59b28c8

ebu_tt_live/node/delay.py
```diff
@@ -25,16 +23,31 @@ def process_document(self, document):
...
+
+        # TODO: add an ebuttm:trace element to the document metadata
```
malikbeytrison: Done in 59b28c8

ebu_tt_live/node/test/test_delay_unit.py
```diff
@@ -1,9 +1,31 @@
...
 
-        pass
+        carriage = MagicMock()
+        reference_clock = MagicMock()
+        delay = 2
+        document_sequence = 'TestSequence'
+        node = RetimingDelayNode('delay_node', carriage, reference_clock, delay, document_sequence)
+        document = MagicMock()
+        node.process_document(document)
+        carriage.emit_document.assert_called_with(document)
```
nigelmegitt: I don't understand what's going on here: does this not require a delay parameter? Sorry this is my ignorance!

ebu_tt_live/node/delay.py
nigelmegitt: Yes, sorry, not sure why I didn't see that.

testing/bdd/features/timing/delay.feature
malikbeytrison: Actually this is what was tested previously, but we changed the code with Zoltan last week to check computed times instead. So, should we also check `begin` and `end` attributes ?

nigelmegitt: Yes, sorry for the confusion, I'm not sure what you discussed with @kozmaz87 exactly, but the behaviour of the `RetimingDelayNode` is defined by the change it makes between inputs and outputs, so that's what the BDD should check, rather than (or as well as) the internal computed values.

malikbeytrison: I see. What we discussed is that when we do `document.validate()`, the computed times should be updated, and if they are updated in a correct way, it means that the specified times have been updated correctly by the `RetimingDelayNode`.

malikbeytrison:
@nigelmegitt To answer your previous summary:  
* unit tests have been added in 69178e0. 
* "BDD tests that check the serialised output": is it the fact to check specified times or yet something else
* sequence ID: done 59b28c8
* `appliedProcessing`: in progress

testing/bdd/features/timing/delay.feature
nigelmegitt: That is technically correct, yes, but I still want to check that the output document is right.

malikbeytrison: Perfect, I'll add the required tests then.

nigelmegitt:
Thanks for the summary @malikbeytrison very helpful. I'm a bit baffled about why coveralls is showing a tiny reduction in coverage - I think it might be a rounding error, looking at the details.

malikbeytrison:
This is surprising for me too, especially given I only removed some code in last commit.

nigelmegitt:
I'm not going to worry about the coverage bit too much when it looks like its a rounding error like this.

nigelmegitt:
Depending on progress with adding `appliedProcessing` (which needs an XSD change) and a decision about `RetimingDelayNode` converting implicitly timed documents to explicitly timed ones, once this has has a code review I would be happy to merge it. 

If we merge before completing the `appliedProcessing` or implicit-> explicit time adjustments then we should log those as issues and add them to the task list for future work. I generally prefer to merge completed bits of functionality sooner rather than later.

ebu_tt_live/node/delay.py
kozmaz87: and now I guess we would like to turn 180 backwards on this with smpte :D

ebu_tt_live/twisted/node.py
```diff
@@ -19,7 +19,9 @@ def __init__(self, consumer, custom_producer):
         self._consumer.registerProducer(self, False)
         self._custom_producer.register_twisted_producer(self)
 
-    def emit_data(self, channel, data):
+    def emit_data(self, channel, data, delay=None):
+        if delay is not None:
+            reactor.callLater(delay, self._consumer.write, channel, data)
         self._consumer.write(channel, data)
```
kozmaz87: Oh my.... I left that bug in there for you to notice...

kozmaz87: The million dollar question: What happens when delay is not None

kozmaz87: This is where unittesting just does not cut it. There was no way you could have tested this with the approach I was suggesting... you should have been able to establish how many calls were made to self._consumer.write, which is 2 n the case of delay which is wrong. This is when twisted's trial comes into play. That is their testing framework to get the reactor to tick in a defined manner. https://twistedmatrix.com/documents/16.4.0/core/howto/testing.html

In particular:

If your code uses Deferreds or depends on the reactor running, you can return a Deferred from your test method, setUp, or tearDown and Trial will do the right thing. That is, it will run the reactor for you until the Deferred has triggered and its callbacks have been run. Don’t use reactor.run() , reactor.stop() , reactor.crash() or reactor.iterate() in your tests.

testing/bdd/features/timing/delay.feature
```diff
@@ -0,0 +1,125 @@
...
+    | 00:00:05   | 00:00:20     | 00:00:03  |          | 00:00:04 |          | 00:00:05 | 00:00:10           | 00:00:25         | 00:00:13          | 00:00:25        | 00:00:17        | 00:00:25      |            |          |          |
+    | 00:00:10   | 00:00:10.500 |           |          | 00:00:03 | 00:00:06 | 01:00:00 | 01:00:10           | 01:00:10.500     | 01:00:13          | 01:00:10.500    | 01:00:13        | 01:00:10.500  |            |          |          |
+    |            |              | 00:00:10  | 00:00:20 | 00:00:05 | 00:00:07 | 00:00:01 | 00:00:11           | 00:00:21         | 00:00:11          | 00:00:21        | 00:00:16        | 00:00:18      |            |          |          |
+  # |            |              |           | 00:00:20 |          | 00:00:10 | 00:00:15 |                    |                  |                   | 00:00:35        |                 | 00:00:25      |            |          |          |
+  # |            |              | 00:00:20  |          |          | 00:00:10 | 00:00:15 | 00:00:35           |                  | 00:00:35          |                 | 00:00:35        | 00:00:10      |            |          |          |
```
nigelmegitt: I'm looking again at this, and not sure why this last line is commented. The others are obviously there because we have not done #291 yet, but this seems to be a different case. Why does the following not apply?

* `updated_body_end` == `updated_div_end` == `updated_p_end` == `00:00:45`

These are the values I think I'd expect for computed times for these input values.

testing/bdd/features/timing/delay.feature
```diff
@@ -0,0 +1,125 @@
...
+    | 00:00:10.0 | 00:00:02.0 | 00:00:12.0         |
+    | 00:00:10.0 | 00:00:00.0 | 00:00:10.0         |
+    # Negative value should really throw an exception
+    | 00:00:10.0 | -00:00:00  | 00:00:10.0         |  
+
+#TODO: test that the order of documents in the input and output sequence is the same (FIFO) 
+
+
+  #TBC: this is a change to the spec (passive delay node)
+  @skip
```
nigelmegitt: Are we skipping these because we have no way to test a buffer delay node?

testing/bdd/features/timing/delay.feature
malikbeytrison: Yes, your reasoning is correct. This is not the same case, as we have a `div_begin` here. Uncommented in  e5c943f

testing/bdd/features/timing/delay.feature
malikbeytrison: Yes

nigelmegitt: Right, I see, that's issue #293 . Thanks.

PR #249, opened 2016-10-25
timeline inconsistency false alarm

kozmaz87:
This is caused by the very unlikely scenario of an exact override of the same document with a newer version of itself... The timeline code did not assume such a precise scenario.


kozmaz87:
I assign it a line above that

nigelmegitt:
Oh yes, silly me!

PR #251, opened 2016-10-25
NotComplexContentError unhandled in segmenter

kozmaz87:
This solution is not as nice as I would like it to be but no time now to make it nicer.


nigelmegitt:
Looks good to me, will merge.

PR #253, opened 2016-10-25
Styling deconflicts its chained style ids properly now...

kozmaz87:
It was mapped wrongly to the source instance instead of the copied one...

### DIFF
--- a/ebu_tt_live/bindings/__init__.py
+++ b/ebu_tt_live/bindings/__init__.py
@@ -756,9 +756,10 @@ class styling(SemanticValidationMixin, raw.styling):
                     for item in self.orderedContent()
                     if isinstance(item, ElementContent) and isinstance(item.value, style_type)
                 ]:
+            copied_style_elem = dataset['instance_mapping'][style_elem]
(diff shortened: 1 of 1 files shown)


ebu_tt_live/bindings/__init__.py
```diff
@@ -756,9 +756,12 @@ def _semantic_after_subtree_copy(self, copied_instance, dataset, element_content
...
                 ]:
-            style_elem_styles = style_elem._semantic_deconflicted_ids(attr_name='style', dataset=dataset)
-            if style_elem_styles:
-                style_elem.style = style_elem_styles
+            copied_style_elem = dataset['instance_mapping'].get(style_elem)
+            # The style may not have been copied at all because it isn't used in the requested segment
+            if copied_style_elem is not None:
+                style_elem_styles = style_elem._semantic_deconflicted_ids(attr_name='style', dataset=dataset)
+                if style_elem_styles:
+                    copied_style_elem.style = style_elem_styles
```
nigelmegitt: Sorry if this is a silly question but what happens to copied_style_elem? I can't see any other references to it in the code. Is it just a reference to another thing rather than a copied object?

Issue #293, opened 2016-12-01
[DELAY NODE] BDD tests for BufferDelayNode


malikbeytrison:
Approach: 

Producer Node -> [1. Consumer Node A that logs resolved document begin time; 2. BufferDelay Node -> Consumer Node B]. Difference between resolved document begin times logged by A and B should be equal to delay period.

nigelmegitt:
I've just realised that the test for passive nodes (as the Buffer Delay Node is) not modifying documents is partially implemented already in #202, which is not yet merged. There may be something we can reuse there...

PR #301, opened 2016-12-19
Malik/issue 291 implicit explicit conversion

malikbeytrison:


ebu_tt_live/node/delay.py
```diff
@@ -76,6 +83,29 @@ def update_children_timing(element, timebase, delay_int):
                 update_children_timing(child.value, timebase, delay_int)
 
 
+def update_body_timing(body, timebase, delay_int):
+
```
nigelmegitt: since this is supposed to add a body begin time maybe it should:

1. be called `insert_body_timing` or something similar
2. have an assertion at the beginning to check that there is in fact no `begin` attribute on the `body` already

ebu_tt_live/node/delay.py
```diff
@@ -93,3 +123,62 @@ def is_explicitly_timed(element):
...
+    def __init__(self, root_element):
+        self._timed_element_stack = []
+        super(UntimedPathFinder, self).__init__(
+            root_element,
+            filter=lambda value, element: isinstance(value, TimingValidationMixin)
+        )
+
+    def _is_begin_timed(self, value):
+        if value.begin is not None:
+            print '{} begin={}'.format(value, value.begin)
```
nigelmegitt: Probably don't need the print for the merge version? Looks like a debug statement unless I've misunderstood.

ebu_tt_live/node/delay.py
```diff
@@ -93,3 +123,62 @@ def is_explicitly_timed(element):
...
+            return True
+        else:
+            return False
+
+    def _before_element(self, value, element=None, parent_binding=None, **kwargs):
+        if self._path_found is True:
+            raise StopBranchIteration()
+        if self._is_begin_timed(value=value):
+            print 'adding {} to stack'.format(value)
```
nigelmegitt: debug statement to remove?

ebu_tt_live/node/delay.py
```diff
@@ -93,3 +123,62 @@ def is_explicitly_timed(element):
...
+            self._timed_element_stack.append(value)
+
+    def _after_element(self, value, element=None, parent_binding=None, **kwargs):
+        if self._is_begin_timed(value=value):
+            bla = self._timed_element_stack.pop()
+            print 'popping {} from stack'.format(bla)
```
nigelmegitt: debug statement to remove?

ebu_tt_live/node/delay.py
```diff
@@ -93,3 +123,62 @@ def is_explicitly_timed(element):
...
+
+    def _process_element(self, value, element=None, parent_binding=None, **kwargs):
+        print value
+        if value.is_timed_leaf() and not len(self._timed_element_stack):
+            print 'path found'
```
nigelmegitt: debug statement to remove?

ebu_tt_live/node/delay.py
```diff
@@ -93,3 +123,62 @@ def is_explicitly_timed(element):
...
+            print 'adding {} to stack'.format(value)
+            self._timed_element_stack.append(value)
+
+    def _after_element(self, value, element=None, parent_binding=None, **kwargs):
+        if self._is_begin_timed(value=value):
+            bla = self._timed_element_stack.pop()
+            print 'popping {} from stack'.format(bla)
+
+    def _process_element(self, value, element=None, parent_binding=None, **kwargs):
+        print value
```
nigelmegitt: Debug statement to remove?

ebu_tt_live/node/delay.py
malikbeytrison: Done in 2d2ffb1.

ebu_tt_live/node/delay.py
malikbeytrison: Done in 2d2ffb1.

ebu_tt_live/node/delay.py
malikbeytrison: Done in 2d2ffb1.

ebu_tt_live/node/delay.py
malikbeytrison: Done in 2d2ffb1.

ebu_tt_live/node/delay.py
malikbeytrison: Done in 2d2ffb1.

ebu_tt_live/node/delay.py
malikbeytrison: 1. I don't 100% agree. I agree to say that we always _insert_ a body begin time, but sometimes, if the end attribute is already present, we _update_ it.

2. I struggled a bit with how to formulate this one. In fact, we could have a body begin time, but if it's the case, it must be None. This is how I did the assertion in 2d2ffb1.

malikbeytrison:
@nigelmegitt here is the summary on your comments:

- print statements removed
- change the naming of the function: not done, see comment above.
- assertion done
- BDD: I uncommented 4 line in 011b923.

kozmaz87:
There is one question here though: What about those skipped tests?

  @skip
  Scenario: Implicitly timed document delay

  #TBC: this is a change to the spec (passive delay node)
  @skip
  Scenario: Implicitly timed document unchanged sequence identifier 

Are these comments still relevant?

malikbeytrison:
@kozmaz87 the first scenario should be called "BufferDelayNode" or something along those lines. I think to distinguish by "explicit" or "implicit" does not make sense anymore as we process the documents regardless of their "type". Also it is planned in #293. The same comment applies for the last two scenarios where we should say "RetimingDelayNode" instead of "Explicitly timed document".

The second scenario was written by @EyalLavi, and I'm not sure about its status now.

testing/bdd/features/timing/delay.feature
```diff
@@ -20,29 +20,29 @@ Feature: Delay of a document sequence
     # Negative value should really throw an exception
     | 00:00:10.0 | -00:00:00  | 00:00:10.0         |  
 
-#TODO: test that the order of documents in the input and output sequence is the same (FIFO) 
+#TODO: test that the order of documents in the input and output sequence is the same (FIFO)
 
 
   #TBC: this is a change to the spec (passive delay node)
   @skip
```
nigelmegitt: We do need to find a way to unskip this, since it is a requirement - @EyalLavi will add the conformance requirement number relationship.

testing/bdd/features/timing/delay.feature
```diff
@@ -20,29 +20,29 @@ Feature: Delay of a document sequence
...
-    | sequence_id_1 | delay      | sequence_id_2 |  
-    | 1             | 00:00:02.0 | 1             |  
-    | xxx           | 00:00:00   | xxx           |  
-    | 99999999999   | 99:00:00   | 99999999999   | 
+    | sequence_id_1 | delay      | sequence_id_2 |
+    | 1             | 00:00:02.0 | 1             |
+    | xxx           | 00:00:00   | xxx           |
+    | 99999999999   | 99:00:00   | 99999999999   |
     # No document generated for negative delay
     | 1             | -00:00:01  |               |
```
nigelmegitt: This last line doesn't feel right here. It's checking a different conformance requirement that may not actually exist.

kozmaz87: It is indeed in an odd place to check for the negative delay things

testing/bdd/features/timing/delay.feature
malikbeytrison: So I can see several things here: 
- the two skipped scenarios. Should we leave them for #293 ?
- the TODO: we implement it for this PR ?
- the line about negative delay: we remove it ?

nigelmegitt: I have the impression that this PR resolves #293 also, so we should unskip them and check they pass, i.e. implement in this PR if possible.

Yes, remove the line about negative delay.

malikbeytrison: Tests with negative delays have been removed. The skipped tests don't pass, because they haven't been implemented for BufferDelayNode.

nigelmegitt:
Just to clarify discussion, although this PR slightly touches (by renaming) BDDs for Buffer Delay Node, it doesn't do anything else, so we will deal with those in a separate PR to address #293.

PR #302, opened 2016-12-21
update licence

frans-ebu:


kozmaz87:
ACK. 
and oops on our part :)

PR #336, opened 2017-02-06
Iss297 R107 discard duplicate sequence pair

EyalLavi:
Scenarios for R107 and 108


testing/bdd/features/segmentation/duplicate_sequence_id+nun.feature
```diff
@@ -0,0 +1,37 @@
...
+    | xml_file            |   
+    | sequence_id_num.xml |   
+    
+
+  # SPEC-CONFORMANCE: R107
+  Scenario: Discard document
+    Given an xml file <xml_file>
+    And it has sequence identifier <seq_id>
+    And it has sequence number <seq_n>
+    When a previous document exists with sequence identifier <prev_seq_id> and sequence number <prev_seq_n>
```
nigelmegitt: I think it makes more sense here to say that when the document is made available and then another document with the same id and number becomes available then when trying to add the second document it is not added. This BDD case never creates the first document, so all the work has to be done in "When a previous document exists..." which presumably creates the first document and adds it?

testing/bdd/features/segmentation/duplicate_sequence_id+nun.feature
```diff
@@ -0,0 +1,37 @@
...
+    # Leading zeros and + not permitted for xs:positiveInteger so there's not much else we can test here.
+
+  # SPEC-CONFORMANCE: R108
+  Scenario: Availability time unchanged when discarding
+    Given an xml file <xml_file>
+    And it has sequence identifier <seq_id>
+    And it has sequence number <seq_n>
+    When a previous document exists with sequence identifier <prev_seq_id> and sequence number <prev_seq_n> 
+    And the previous document has availability time <prev_avail>
+    Then the previous document has availability time <post_avail>
```
nigelmegitt: I would refactor this one too - given a file with sequence id and number and it is added to the sequence with availability time `<prev_avail>` and an attempt is made to add a new file with same sequence id and number to the sequence then the availability time of the file in the sequence with that sequence id and number is `<prev_avail>`.

testing/bdd/features/segmentation/duplicate_sequence_id+nun.feature
```diff
@@ -10,28 +10,50 @@ Feature: Discard document with already seen pair of sequence identifier and sequ
...
-    And the previous document has availability time <prev_avail>
-    Then the previous document has availability time <post_avail>
+    And it has sequence identifier <seq_id_1>
+    And it has sequence number <seq_n_1>
+    And the document is added to the sequence
+    And the document has availability time <avail_time_1>    
+    When a document arrives 
+    And it has sequence identifier <seq_id_2> 
+    And it has sequence number <seq_n_2> 
+    Then the previous document has availability time <avail_time_2>
```
nigelmegitt: Surely avail_time_1 and avail_time_2 are always the same in this case? What is the case where they would differ?

nigelmegitt: Just wondering: "previous" -> "first" for more clarity ?

testing/bdd/features/segmentation/duplicate_sequence_id+nun.feature
```diff
@@ -10,28 +10,50 @@ Feature: Discard document with already seen pair of sequence identifier and sequ
...
 
     Examples:
-    | seq_id | seq_n | prev_seq_id | prev_seq_n | prev_avail   | post_avail   |  
-    | n      | 1     | n           | 1          | 00:00:00.000 | 00:00:00.000 | 
-     
-
+    | seq_id_1 | seq_n_1 | avail_time_1 | seq_id_2 | seq_n_2      | avail_time_2 |  
+    | n        | 1       | n            | 1        | 00:00:00.000 | 00:00:00.000 |
```
nigelmegitt: What is `n` for an availability time?

PR #337, opened 2017-02-07
R12 BDD test for buffer delay node

kozmaz87:
This solves #338


nigelmegitt:
Thanks @kozmaz87 please could you add the issue that this resolves or addresses to the description of this pull request?

Issue #308, opened 2017-01-17
[Websocket] Escape sequenceIdentifier when using in a URL


nigelmegitt:
Since `ebuttp:sequenceIdentifier` is an `xs:string` it can contain characters that need to be escaped in URLs. Escape this path segment and de-escape it as needed.

Possibly helpful documentation at https://twistedmatrix.com/documents/16.3.1/api/twisted.python.url.URL.html for doing this in twisted.

nigelmegitt:
@kozmaz87 is this actually done? There's no PR referencing the issue but I think we did implement it...

kozmaz87:
7e637b6  - f904d97 commits add in the unittesting capability and the functionality. The rest are preparation for the full system testing of the same feature.

nigelmegitt:
This is close enough for now, we should have a different issue to complete the WebSocket BDD testing if we need to do that.

PR #354, opened 2017-02-16
#308 issue system test bits

kozmaz87:
this is currently skipped.

[additional notes added by @nigelmegitt later]

Partial implementation to support testing of WebSocket, not very closely tied to #308, but worth keeping. The problem here is that in the test framework if we run a reactor we have no way to get control back and stop it. This will take further work.


nigelmegitt:
Merging this skipped code, useful to pick up later if we want to BDD test WebSocket, and to remind us how to get an ephemeral port on an unknown machine and use it in the config script.

PR #355, opened 2017-02-16
Task/29

kozmaz87:
This is what it took to add distributor node.


nigelmegitt:
Issue #357 raised for the problem mentioned above.

nigelmegitt:
By the way I would be happy to merge this as is but will wait in case a fix lands for #357 on this PR.

nigelmegitt:
By the way I verified the fix for #357.

PR #365, opened 2017-03-06
hotfix: missing things for simple filesystem exporter and ebuttd encoder

kozmaz87:
This is release 2 fixes and they should be merged back into master. Don't remove the branch in the end.


nigelmegitt:
Merging now, we'll need an issue to unit test this since it should have been caught previously.

PR #376, opened 2017-03-16
Hamees01 test ui issue370

SKHameed86:
Check changes made to test.html


setup.py
```diff
@@ -23,7 +23,7 @@
 
 setup(
     name="ebu-tt-live",
-    version="2.0",
+    version="3.0",
```
nigelmegitt: Should this change be in here? @kozmaz87 please advise.

nigelmegitt:
Reverted `setup.py` version update which should not have been merged into this branch (since we are targeting release/2.0 not release/3.0). Will merge.

PR #384, opened 2017-03-27
Refactored clock code

SKHameed86:


ebu_tt_live/ui/user_input_producer/user_input_producer.html
```diff
@@ -4,6 +4,7 @@
       <meta charset="UTF-8"></meta>
       <script src="https://code.jquery.com/jquery-3.1.0.min.js"></script>
       <script src="template/user_input_producer_template.js"></script>
+      <script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.18.1/moment.min.js"></script>
```
nigelmegitt: Can we bring this in locally please?

ebu_tt_live/ui/user_input_producer/uip.js
```diff
@@ -423,8 +385,10 @@ $(document).ready(function() {
     $("#scheduled-send-clock-selector").change(handleScheduledSendSetupDependingOptions);
 
     $("#synchronize-media-clock-button").click(function() {
-        scheduled_send_media_clock_offset = Date.now();
-        stopResetRunningClock();
+        scheduled_send_media_clock_offset = moment();
+        //stopResetRunningClock();
+        $('#running-clock').hide();
+        $('#running-clock-media').show();
         interval_running_clock = setInterval(updateRunningClockMedia, 500);
```
nigelmegitt: Why do we need both a `setInterval()` _and_ a `setTimeout()` (lines 116 and 123)? Seems like we should need one or the other but not both... Also, it looks as though every time the `synchronize-media-clock-button` is clicked another interval is created regardless of whether or not one already existed. I think this is leaky and probably consumes resources unnecessarily, especially because the current interval is not cleared (now that line 389 is commented out). Or possibly I have not understood...

SKHameed86: On Friday, Zoltan explained to me that because there were too many `setInterval` instances, it was causing the browser to crash, so `setTimeout` was the better choice for the clocks.

I wasn't 100% sure what line 389 was doing, so I commented it out for that instance - I can put it back.

ebu_tt_live/ui/user_input_producer/user_input_producer.html
SKHameed86: How do I do that?

ebu_tt_live/ui/user_input_producer/uip.js
nigelmegitt: OK well one thing that the function in line 389 does is to clear the current interval before creating a new one!

ebu_tt_live/ui/user_input_producer/uip.js
```diff
@@ -27,7 +27,7 @@ $(document).ready(function() {
     /********************************************************** Options to show or not depending on other options and general purpose functions****************************/
 
     function stopResetRunningClock() {
-        $("#running-media-clock").html("");
+        //$("#running-clock").empty();
```
kozmaz87: What is the deal about this line?

PR #387, opened 2017-03-28
Feature/live message

kozmaz87:
Live message use cases on UIP work to a certain extent


ebu_tt_live/node/distributing.py
```diff
@@ -21,16 +21,19 @@ def __init__(self, node_id, producer_carriage=None, consumer_carriage=None, **kw
         )
 
     def process_document(self, document, raw_xml=None, **kwargs):
-        if self.check_document(document) is True:
-            if raw_xml is not None:
-                data = raw_xml
+        if self.is_document(document):
```
nigelmegitt: This looks like another change that needs to be moved into release/3.0 separately, ideally.

ebu_tt_live/node/handover.py
```diff
@@ -60,23 +60,31 @@ def process_document(self, document, **kwargs):
...
+        if self.is_document(document):
+            if self.check_if_document_seen(document=document) is True \
+                    and document.authors_group_identifier == self._authors_group_identifier \
+                    and document.authors_group_control_token is not None:
+                if self._current_token is None or self._current_token < document.authors_group_control_token:
+                    # Switch input
+                    self._current_selected_input_sequence_id = document.sequence_identifier
+                    emit = True
+                elif self._current_selected_input_sequence_id == document.sequence_identifier:
+                    emit = True
```
nigelmegitt: Logic looks correct.

ebu_tt_live/node/handover.py
```diff
@@ -60,23 +60,31 @@ def process_document(self, document, **kwargs):
...
+
+            if emit is True:
+                # Update token
+                self._current_token = document.authors_group_control_token  # So are we going to error here?
```
nigelmegitt: What's happening with this comment? Looks like we cannot get here if the  authors group control token is absent. Or is there another case to consider?

ebu_tt_live/xsd/ebutt_livemessage.xsd
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           xmlns="http://www.w3.org/1999/xhtml"
```
nigelmegitt: xhtml? Is that correct? Seems surprising not to use the XMLSchema namespace as the default and remove all the `xs:`s everywhere.

ebu_tt_live/xsd/ebutt_livemessage.xsd
```diff
@@ -0,0 +1,42 @@
...
+           xmlns:ebuttp="urn:ebu:tt:parameters"
+           xmlns:ebuttlm="urn:ebu:tt:livemessage"
+           targetNamespace="urn:ebu:tt:livemessage" elementFormDefault="qualified"
+           xml:lang="en">
+    <xs:import namespace="urn:ebu:tt:parameters" schemaLocation="ebutt_parameters.xsd"/>
+
+    <xs:complexType name="message_header_type">
+        <xs:annotation>
+            <xs:documentation>Message header to help a recipient identify the purpose of the message
+            and decide whether it is relevant to them and if so what type of payload it contai if any
```
nigelmegitt: Typo: s/contai/contains

ebu_tt_live/xsd/ebutt_livemessage.xsd
```diff
@@ -0,0 +1,42 @@
...
+            <xs:element name="sender" type="xs:string" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="recipient" type="xs:string" minOccurs="0" maxOccurs="unbounded"/>
+            <xs:element name="type" type="xs:NCName" />
+        </xs:sequence>
+    </xs:complexType>
+    
+    <xs:complexType name="message_type">
+        <xs:annotation>
+            <xs:documentation>The message container type that can be used as a root document element
+            to create a new message. It shall specify its header and it could speficy a payload.
```
nigelmegitt: spelling/typo: s/speficy/specify

ebu_tt_live/xsd/ebutt_metadata.xsd
```diff
@@ -526,14 +526,18 @@ Please note that the EBU-TT XML Schema is a helping document and NOT normative b
...
+	<xs:attribute name="authorsGroupSelectedSequenceIdentifier">
 		<xs:annotation>
 			<xs:documentation>
-				Allows an author to signal their intent to continue providing a stream, or to request that another author begins creating
-				output with a higher ebuttp:authorsGroupControlToken value.
+				The selected sequenceIdentifier is to be moved here by the handover node.
 			</xs:documentation>
 		</xs:annotation>
+		<xs:simpleType>
+            <xs:restriction base="xs:string">
```
nigelmegitt: Nice catch!

test.html
```diff
@@ -15,7 +15,7 @@
                return;
             }
 
-            wsuri = host_input.value
+            wsuri = host_input.value + '/' + channel_input.value + '/subscribe';
```
nigelmegitt: Seems like the changes to this file shouldn't be part of this pull request...

ebu_tt_live/carriage/test/test_filesystem.py
```diff
@@ -48,11 +49,86 @@ def test_emit_document(self):
...
+        # An acceptable workaround could be to take the the last received availability time - local clock
+        # offset for the sequence and compute an estimate using the local clock
+        #
+        # Since this was not a requirement the message gets discarded...
+        #
+        # There are also an unsettling number of odd defined/not defined combinations of arguments at this
+        # level, none of which is specified to trigger a particular response: such as receiving
+        # availability_time but no time_base...etc.
+
+        node = MagicMock(spec=IProducerNode)
```
kozmaz87: Yes looks like a nasty copypaste... and could indeed be saved on. Since this is a test it is not that paramount. I was admittedly a bit lazy as I started working out these quirky cases...

On the topic of mocking. When it comes to unittesting something 'mocking' means that you replace the outside dependency with some simplified dummy object that would detach you from issues/configuration requirements of other components so that your testing can focus on the unit at hand without noise.

The 1st line defines a node that implements the IProducerNode interface and the 2nd tells it that if its provides function is called it should return six.text_type... This is necessary because otherwise the carriage mechanism would not be compatible with the node when the carriage mechanism does its compatibility checks.

SKHameed86: Ah...okay. Thank you for the clarification! : )

SKHameed86:
It seems to be okay to me - a few copy/paste instances, but as Zoltan explained...it is a test/draft.

ebu_tt_live/node/distributing.py
nigelmegitt: I think this comment is no longer applicable since I merged Release/3.0 into this branch.

PR #395, opened 2017-03-30
iss392 scenarios for handover reqs

### LINKED ISSUE #392: [HANDOVER] Write BDDs for new requirements
See #390.

EyalLavi:
Fixes #392


testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,84 @@
...
+
+  # SPEC-CONFORMANCE: R23, R25, 120
+  Scenario: Switch to sequence with higher value token
+    Given an xml file <xml_file>
+    And a handover node
+    When it has sequence identifier <seq_id_1>
+    And it has authorsGroupControlToken <token_1>
+    And the document is generated
+    And the document is processed 
+    # ^^^ ASSUMING THIS STEP COVERS CACHED TOKENS, IE THE SECOND DOCUMENTS DOESN'T HAVE TO FOLLOW IMMEDIATELY
```
nigelmegitt: We probably don't need this comment?

testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,84 @@
...
+    And it has authorsGroupControlToken <token_2>
+    And the document is generated
+    And the document is processed
+    Then the emitted document has <authorsGroupSelectedSequenceIdentifier> 
+
+    Examples:
+    | seq_id_1   | token_1         | seq_id_2   | token_2         | authorsGroupSelectedSequenceIdentifier |  
+    | sequence_a | 12345           | sequence_b | 12346           | sequence_b                             |  
+    | sequence_a | 1               | sequence_b | 0               | sequence_a                             |  
+    | sequence_a | 0               | sequence_b |                 | sequence_a                             |
```
nigelmegitt: 0 is not a permitted value so sequence_a should not be selected.

testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,84 @@
...
+    And the document is generated
+    And the document is processed
+    Then the emitted document has <authorsGroupSelectedSequenceIdentifier> 
+
+    Examples:
+    | seq_id_1   | token_1         | seq_id_2   | token_2         | authorsGroupSelectedSequenceIdentifier |  
+    | sequence_a | 12345           | sequence_b | 12346           | sequence_b                             |  
+    | sequence_a | 1               | sequence_b | 0               | sequence_a                             |  
+    | sequence_a | 0               | sequence_b |                 | sequence_a                             |  
+    | sequence_a | 999999999999999 | sequence_b | 999999999999998 | sequence_a                             |
```
nigelmegitt: How many 9s is that? A lot!

nigelmegitt: This 15 digit number > sys.maxint on 32 bit Python 2.7 so it will probably fail on some machines (it is platform dependent). We can reduce it to a smaller value, e.g. 9 digits.

testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,84 @@
...
+    And the document is processed
+    Then the emitted document has <authorsGroupSelectedSequenceIdentifier> 
+
+    Examples:
+    | seq_id_1   | token_1         | seq_id_2   | token_2         | authorsGroupSelectedSequenceIdentifier |  
+    | sequence_a | 12345           | sequence_b | 12346           | sequence_b                             |  
+    | sequence_a | 1               | sequence_b | 0               | sequence_a                             |  
+    | sequence_a | 0               | sequence_b |                 | sequence_a                             |  
+    | sequence_a | 999999999999999 | sequence_b | 999999999999998 | sequence_a                             |  
+
```
nigelmegitt: We should have an additional line to test that sequence_a token 3 followed by sequence_a token 2 results in sequence_a still. And maybe check that in this case the second document is actually emitted.

testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,84 @@
...
+    And it has sequence identifier <seq_id_2>
+    And it has authorsGroupControlToken <token_2>
+    And the document is generated
+    And the document is processed
+    Then the emitted document has <authorsGroupSelectedSequenceIdentifier> 
+
+    Examples:
+    | seq_id_1   | token_1         | seq_id_2   | token_2         | authorsGroupSelectedSequenceIdentifier |  
+    | sequence_a | 12345           | sequence_b | 12346           | sequence_b                             |  
+    | sequence_a | 1               | sequence_b | 0               | sequence_a                             |
```
nigelmegitt: There will be no second document emitted here (even if we make `token_2` valid but less than `token_1`), so which document's `authorsGroupSelectedSequenceIdentifier` is being checked? Might need to check in this case that another document does not arrive, instead.

testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,84 @@
...
+    And it has sequenceNumber 1
+    And it has <authorsGroupIdentifier_2>
+    Then adding doc2 to the sequence results in an error
+
+    Examples:
+    | authorsGroupIdentifier_1 | authorsGroupIdentifier_1 |  
+    | 1                        | 0                        |  
+    | 0                        | 1                        |  
+    |                          | 1                        |  
+    # ^^^ Assuming NULL means authordGroupIdentifier is present but empty (if the element doesn't exist the document is valid!)
```
nigelmegitt: Agree with this comment!

testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,84 @@
...
+
```
nigelmegitt: I think there should be a valid case too where the authorsGroupIdentifiers are identical and that does not cause an error.

testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,84 @@
...
+  Scenario: Missing group identifiers in a sequence 
+    Given a test sequence
+    And an xml file <xml_file>
+    When it has sequenceNumber 1
+    And it has <authorsGroupIdentifier_1>
+    And doc1 is added to the sequence
+    And we create a new document
+    And it has sequenceNumber 1
+    And it does not have element authorsGroupIdentifier
```
nigelmegitt: It is an attribute not an element.

testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,84 @@
...
+    # ^^^ Assuming NULL means authordGroupIdentifier is present but empty (if the element doesn't exist the document is valid!)
+
+  Scenario: Missing group identifiers in a sequence 
+    Given a test sequence
+    And an xml file <xml_file>
+    When it has sequenceNumber 1
+    And it has <authorsGroupIdentifier_1>
+    And doc1 is added to the sequence
+    And we create a new document
+    And it has sequenceNumber 1
```
nigelmegitt: Should they have the same `sequenceNumber`? ( @EyalLavi says they should be the same `sequenceIdentifier`)

testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,84 @@
...
+
+  # SPEC-CONFORMANCE: R121
+  # The Handover Manager shall not emit any documents derived from documents that do not contain both the parameters 
+  # ebuttp:authorsGroupIdentifier and ebuttp:authorsGroupControlToken.
+  Scenario: Documents with missing group identifier and token
+    Given an xml file <xml_file>
+    And a handover node
+    When it does not have element authorsGroupControlToken 
+    And it does not have element authorsGroupIdentifier 
+    Then the document is not added to sequence
```
nigelmegitt: What about the other permutations of one present but not the other? There should be 3 "document is not emitted by the Handover Manager" permutations, where either one or both are omitted.

testing/bdd/features/handover/handover.feature
```diff
@@ -13,44 +13,66 @@ Feature: Handover algorithm
...
     And a handover node
     When it has sequence identifier <seq_id_1>
+    And it has authorsGroupIdentifier <group_1>
     And it has authorsGroupControlToken <token_1>
     And the document is generated
     And the document is processed 
-    # ^^^ ASSUMING THIS STEP COVERS CACHED TOKENS, IE THE SECOND DOCUMENTS DOESN'T HAVE TO FOLLOW IMMEDIATELY 
     And another document arrives 
     And it has sequence identifier <seq_id_2>
+    And it has authorsGroupIdentifier <group_1>
```
nigelmegitt: Shouldn't this be `group_2`?

testing/bdd/features/handover/handover.feature
```diff
@@ -13,44 +13,66 @@ Feature: Handover algorithm
...
     Then the emitted document has <authorsGroupSelectedSequenceIdentifier> 
 
     Examples:
-    | seq_id_1   | token_1         | seq_id_2   | token_2         | authorsGroupSelectedSequenceIdentifier |  
-    | sequence_a | 12345           | sequence_b | 12346           | sequence_b                             |  
-    | sequence_a | 1               | sequence_b | 0               | sequence_a                             |  
-    | sequence_a | 0               | sequence_b |                 | sequence_a                             |  
-    | sequence_a | 999999999999999 | sequence_b | 999999999999998 | sequence_a                             |  
+    | seq_id_1   | group_1 | token_1   | seq_id_2   | group_2 | token_2   | authorsGroupSelectedSequenceIdentifier |  
+    | sequence_a | 1       | 12345     | sequence_b | 1       | 12345     | sequence_b                             |
```
nigelmegitt: I think the selected sequence here should be `sequence_b` since the token is the same as the first document, not greater.

testing/bdd/features/handover/handover.feature
```diff
@@ -13,44 +13,66 @@ Feature: Handover algorithm
...
     Examples:
-    | seq_id_1   | token_1         | seq_id_2   | token_2         | authorsGroupSelectedSequenceIdentifier |  
-    | sequence_a | 12345           | sequence_b | 12346           | sequence_b                             |  
-    | sequence_a | 1               | sequence_b | 0               | sequence_a                             |  
-    | sequence_a | 0               | sequence_b |                 | sequence_a                             |  
-    | sequence_a | 999999999999999 | sequence_b | 999999999999998 | sequence_a                             |  
+    | seq_id_1   | group_1 | token_1   | seq_id_2   | group_2 | token_2   | authorsGroupSelectedSequenceIdentifier |  
+    | sequence_a | 1       | 12345     | sequence_b | 1       | 12345     | sequence_b                             |  
+    | sequence_a | a       | 999999999 | sequence_b | a       | 999999999 | sequence_a                             |  
+    | sequence_a | a       |           | sequence_b | a       |           | sequence_b                             |
```
nigelmegitt: In this case no document will be emitted because the token is missing.

testing/bdd/features/handover/handover.feature
```diff
@@ -60,25 +82,36 @@ Feature: Handover algorithm
     And doc1 is added to the sequence
     And we create a new document
     And it has sequenceNumber 1
-    And it does not have element authorsGroupIdentifier
+    And it does not have attribute authorsGroupIdentifier
     Then doc2 is added to the sequence
```
nigelmegitt: is doc2 the new document? Since it has the same `sequenceNumber` it will be ignored and not added to the sequence, regardless of `authorsGroupIdentifier`.

testing/bdd/features/handover/handover.feature
```diff
@@ -60,25 +82,36 @@ Feature: Handover algorithm
...
   # The Handover Manager shall not emit any documents derived from documents that do not contain both the parameters 
   # ebuttp:authorsGroupIdentifier and ebuttp:authorsGroupControlToken.
   Scenario: Documents with missing group identifier and token
     Given an xml file <xml_file>
     And a handover node
-    When it does not have element authorsGroupControlToken 
-    And it does not have element authorsGroupIdentifier 
+    When it does not have attribute authorsGroupControlToken 
+    And it does not have attribute authorsGroupIdentifier 
     Then the document is not added to sequence
```
nigelmegitt: maybe this should say `Then the handover node does not emit a derived document` to be precise?

testing/bdd/features/handover/handover.feature
```diff
@@ -60,25 +82,36 @@ Feature: Handover algorithm
...
 
+  Scenario: Documents with missing group identifier
+    Given an xml file <xml_file>
+    And a handover node
+    When it does not have attribute authorsGroupControlToken 
+    Then the document is not added to sequence
```
nigelmegitt: `Then the handover node does not emit a derived document` may be appropriate here too?

testing/bdd/features/handover/handover.feature
```diff
@@ -60,25 +82,36 @@ Feature: Handover algorithm
...
 
+  Scenario: Documents with missing token
+    Given an xml file <xml_file>
+    And a handover node
+    When it does not have attribute authorsGroupControlToken 
+    Then the document is not added to sequence
```
nigelmegitt: `Then the handover node does not emit a derived document` may be appropriate here too?

testing/bdd/features/handover/handover.feature
EyalLavi: Perhaps but Zoltan asked that I stick to existing step definitions.

testing/bdd/features/handover/handover.feature
EyalLavi: As above.

testing/bdd/features/handover/handover.feature
nigelmegitt: Hmm okay - @kozmaz87 do we have a step definition for a handover node emitting a document?

kozmaz87:
This branch is merged already into the PR branch that implements these scenarios

EyalLavi:
Closing since #406 is more recent and complete.

PR #403, opened 2017-03-31
Fix #399

### LINKED ISSUE #399: [BDD] implement test for empty sequence id is invalid
Recent spec change has made having an empty sequence id invalid. Test for this in the BDD.

nigelmegitt:
Fix #399 but this time based directly on the current release/3.0 branch so this is safer to merge.


nigelmegitt:
This was done with @kozmaz87 and @EyalLavi watching, and is very simple, so when the CI build finishes I will merge directly, assuming it passes as it does on my machine locally.

EyalLavi:
Time for bed!

-----Original Message-----

nigelmegitt:
zzzzz

PR #406, opened 2017-04-05
Task/397

kozmaz87:
This is my take on issue 397. I believe a decent solution needed all these things here. Please review the BDD in full again (both of them) as they significantly changed. They should be creating the same scenarios just the wording changed.


testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,66 @@
...
+    And it has <sequence_identifier2> and <sequence_number2>
+    And it has <authors_group_identifier2>
+    And it has <authors_group_control_token2>
+    And the document is generated
+    Then handover node errors when processing document
+
+    Examples:
+    | sequence_identifier1 | sequence_number1 | authors_group_identifier1 | authors_group_control_token1 | sequence_identifier2 | sequence_number2 | authors_group_identifier2 | authors_group_control_token2 |
+    | seq1                 | 1                | foo                       | 1                            | seq1                 | 1                | bar                       | 2                            |
+
```
EyalLavi: According to R107 the second document will be discarded because it has an identical pair of sequence number and sequence identifier. But here we generate an error. Which action takes precedence?

EyalLavi: Add an example with same sequence identifier and different sequence number.

testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,66 @@
...
+
+  # SPEC-CONFORMANCE: R121
+  # The Handover Manager shall not emit any documents derived from documents that do not contain both the parameters 
+  # ebuttp:authorsGroupIdentifier and ebuttp:authorsGroupControlToken.
+  Scenario: Documents with missing group identifier and token
+    Given a handover node with <authors_group_identifier> and <sequence_identifier>
```
EyalLavi: The authors group identifier and sequence identifiers are properties of documents, not of nodes.

testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,66 @@
...
+    Examples:
+    | xml_file     | sequence_identifier | authors_group_identifier |
+    | handover.xml | handoverOutputSeq   | handoverTest01           |
+
+
+  # SPEC-CONFORMANCE: R24, R122
+  # Within a single sequence, all documents that contain  ebuttp:authorsGroupIdentifier shall have the same ebuttp:authorsGroupIdentifier
+  # The handover node should throw an error when this happens
+  Scenario: Different group identifiers in a sequence 
+    Given a handover node with <authors_group_identifier> and <sequence_identifier>
```
EyalLavi: The authors group identifier and sequence identifiers are properties of documents, not of nodes.

testing/bdd/features/handover/handover_algorithm.feature
```diff
@@ -33,18 +33,17 @@ Feature: Handover algorithm
...
     And the emitted documents have <authors_group_identifier> and they specify a token
 
     Examples:
-    | sequence_identifier1 | sequence_number1 | authors_group_control_token1 | sequence_identifier2 | sequence_number2 | authors_group_control_token2 | emitted_documents |
-    | seq1                 | 1                | 1                            | seq2                 | 1                | 2                            | 2                 |
-    | seq1                 | 1                | 2                            | seq2                 | 1                | 3                            | 2                 |
-    | seq1                 | 1                |                              | seq2                 | 1                | 1                            | 1                 |
+    | sequence_identifier1 | sequence_number1 | authors_group_control_token1 | sequence_identifier2 | sequence_number2 | authors_group_control_token2 | authors_group_selected_sequence_identifiers |
+    | seq1                 | 1                | 1                            | seq2                 | 1                | 2                            | seq1,seq2                                   |
```
EyalLavi: Can authors_group_selected_sequence_identifiers have two comma-separated values?

testing/bdd/test_handover.py
```diff
@@ -79,3 +96,13 @@ def then_handover_parameter_passthrough(given_handover_node, authors_group_ident
     for pos_args, kw_args in given_handover_node.producer_carriage.emit_data.call_args_list:
         assert kw_args['data'].authors_group_identifier == authors_group_identifier
         assert kw_args['data'].authors_group_control_token is not None
+
+
+@then('the emitted documents have <authors_group_selected_sequence_identifiers>')
+def then_authors_group_selected_sequence_id(given_handover_node, authors_group_selected_sequence_identifiers):
+    # NOTE: The comma is a valid sequenceIdentifier character but we use it as a divisor in the test. Make sure
+    # the tests use sequence identifiers without commas.
+    seq_ids = map(lambda x: x.strip(), authors_group_selected_sequence_identifiers.split(','))
```
EyalLavi: See comment above. How can a document have two sources?

testing/bdd/features/handover/handover.feature
nigelmegitt: I don't completely agree with @EyalLavi 's comment above - it is required that a handover manager be configured to deal with a specific authors group identifier and produce output with a particular sequence identifier. Sequence identifiers on incoming documents are distinct though.

testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,66 @@
...
+    And an xml file <xml_file>
+    When it has <sequence_identifier1> and <sequence_number1>
+    And it has <authors_group_identifier1>
```
nigelmegitt: In the table below `authors_group_identifier1` differs from the `authors_group_identifier` defined above so surely in the following steps the handover node does not produce any output based on this input? I think the error, if there is one (as opposed to ignoring the sequence altogether) is generated before line 22 "`And new document is created`".

testing/bdd/features/handover/handover.feature
nigelmegitt: I think this is testing the requirement that authors group identifier, when present, is the same in all documents in the sequence, but it is not clear that the two documents here are both being provided as _inputs_ to the handover manager. I agree that the second doc gets discarded before the error is generated here; however that would easily be fixed by changing the sequence number of the second document to, say, 2.

testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,66 @@
...
+    And it has <authors_group_control_token1>
+    And the document is generated
+    And handover node processes document
+    And new document is created
```
nigelmegitt: Is this a new incoming document to the handover node, or a document that the handover node is generating? We may need to change the wording to clarify the situation here.

testing/bdd/features/handover/handover.feature
```diff
@@ -0,0 +1,66 @@
...
+    And it has <authors_group_control_token1>
+    And the document is generated
+    And handover node processes document
+    Then handover node emits <emitted_documents> documents
+
+    Examples:
+    | sequence_identifier1 | sequence_number1 | authors_group_identifier1 | authors_group_control_token1 | emitted_documents |
+    | seq1                 | 1                |                           | 10                           | 0                 |
+    | seq1                 | 1                | foo                       |                              | 0                 |
+    | seq1                 | 1                |                           |                              | 0                 |
```
nigelmegitt: Handover node will not emit anything in any of these cases because `authors_group_identifier1` != `authors_group_identifier`. Suggest adding a line like the second one where `foo` is replaced by `handoverTest01` - if `authors_group_control_token1` is still empty then 0 documents should be emitted.

testing/bdd/features/handover/handover.feature
EyalLavi: We've been using consistent wording to minimise the number step definitions. I'm not sure of the  consequences.

tomrosier:
No comments to add. I would agree with the comments that have been made.

EyalLavi:
There was some confusion regarding overlaps between this PR and PR #395. This PR is more up to date and it contains tests for all the requirements so we can close #395 without merging.

nigelmegitt:
Just wanted to add an update because this PR has been open for a while. The spec-conformance requirements are changing a little so this may need some tweaking. The effective algorithm isn't changing though.

testing/bdd/features/handover/handover.feature
nigelmegitt: Just to note that those two deleted (failing) test cases were correctly not failing. R24, as a reminder, says:

> Within a single sequence, all documents that contain the element `ebuttp:authorsGroupIdentifier` shall have the same `ebuttp:authorsGroupIdentifier`.

The lines were generating two documents in a sequence one that contains the attribute and the other that does not. That doesn't and shouldn't be throwing an error.

In fact there is a separate test later for documents without a group identifier attribute causing no document to be emitted, so the non-emission of a document without a group id is already tested.

testing/bdd/features/handover/handover.feature
nigelmegitt: I've looked at what this line does, and it actually clears the template dict so that a new document can be created. So it does kind of say what it means. I'm able to live with this, even though I think it is a bit confusing to the newcomer.

testing/bdd/features/handover/handover.feature
nigelmegitt: This comment has been addressed with the following line, which was added later (but did not "outdate" this comment because it did not affect the line it was on).

testing/bdd/features/handover/handover_algorithm.feature
nigelmegitt: I would like to add back in the `emitted_documents` count test. I will do that. There are also some other test cases that can be added.

nigelmegitt:
All review comments have been addressed, will merge.

Issue #326, opened 2017-01-26
simple-consumer does not clear state on input connections being closed


nigelmegitt:
The simple consumer should reset its state when all input connections are closed so that a new connection with a different sequence id can be created and it will continue to work. Currently it sets up with the first provided sequence id, then if a new connection is made with a different sequence id, it produces the following (verbose) output:

```shell
[INFO] (2017-01-26 17:42:14,759) in document_logger[644] - Document abc/def__35 inserted into sequence
[INFO] (2017-01-26 17:42:14,759) in document_logger[649] - New document inserted:
p_type Timings: [(None - None)(0:00:00 - 0:00:05)(0:00:00 - 0:00:05)]
as akljsd
<br />
auh asj
[INFO] (2017-01-26 17:43:44,966) in document_logger[44] - Document clock_utc_seq__1 received. Calculated activation: [0:00:00; 0:00:05]
Unhandled Error
Traceback (most recent call last):
  File "/Users/megitn02/Code/ebu/ebu-tt-live-toolkit/venv/lib/python2.7/site-packages/twisted/python/log.py", line 103, in callWithLogger
    return callWithContext({"system": lp}, func, *args, **kw)
  File "/Users/megitn02/Code/ebu/ebu-tt-live-toolkit/venv/lib/python2.7/site-packages/twisted/python/log.py", line 86, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "/Users/megitn02/Code/ebu/ebu-tt-live-toolkit/venv/lib/python2.7/site-packages/twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/Users/megitn02/Code/ebu/ebu-tt-live-toolkit/venv/lib/python2.7/site-packages/twisted/python/context.py", line 81, in callWithContext
    return func(*args,**kw)
--- <exception caught here> ---
  File "/Users/megitn02/Code/ebu/ebu-tt-live-toolkit/venv/lib/python2.7/site-packages/twisted/internet/selectreactor.py", line 149, in _doReadOrWrite
    why = getattr(selectable, method)()
  File "/Users/megitn02/Code/ebu/ebu-tt-live-toolkit/venv/lib/python2.7/site-packages/twisted/internet/tcp.py", line 208, in doRead
    return self._dataReceived(data)
[...]

PR #431, opened 2017-05-11
Issue 360 tom rosier

tomrosier:


ebu_tt_live/examples/config/retiming_delay_issue_360_example.json
```diff
@@ -0,0 +1,61 @@
+{
```
nigelmegitt: We probably don't want to keep the issue-demonstrating JSON as part of the issue fix, since it is a weird configuration case. If we want to test that the fix is not reintroduced this should be done via unit tests or BDDs. If I don't remember, remind me to remove it before merging.

nigelmegitt:
It would be helpful to apply this fix also to EBUTTDEncoder and SimpleConsumer - the latter is #326. Then we would have the question whether to factor the code out into something like a `check_if_can_handle_sequence(self, sequence_identifier)` method in `AbstractConsumerNode` which would be called from the `process_document` method. Or alternatively this could be a mix-in for those nodes that have specific "only one sequence" requirements.

tomrosier:
Latest commit should also fix issue #326

ebu_tt_live/node/base.py
```diff
@@ -82,6 +83,13 @@ def is_document(self, document):
         else:
             return False
 
+    def has_document_sequence_identifier_used(self, document):
```
nigelmegitt: I'm having trouble understanding what this function might do from its name. Maybe something like `limit_sequence_to_one` would be better? This is so that it is obvious that we should only call it when we actually want to limit the number of input sequences. Happy for other ideas, but it definitely needs to change.

PR #435, opened 2017-05-16
Issue 292 tom rosier

tomrosier:


ebu_tt_live/node/delay.py
```diff
@@ -2,8 +2,9 @@
 import six
 import logging
 from .base import AbstractCombinedNode
-from datetime import timedelta
+from datetime import timedelta, datetime
 from ebu_tt_live.bindings._ebuttdt import LimitedClockTimingType, FullClockTimingType
+from ebu_tt_live.bindings import _ebuttm as metadata, TimingValidationMixin
```
nigelmegitt: Why do you need TimingValidationMixin?

ebu_tt_live/xsd/ebutt_metadata.xsd
```diff
@@ -413,28 +413,22 @@ Please note that the EBU-TT XML Schema is a helping document and NOT normative b
...
-						<xs:sequence>
-							<xs:any namespace="##other" processContents="lax" minOccurs="0" maxOccurs="unbounded"/>
-						</xs:sequence>
-						<xs:attribute name="process" type="xs:string" use="required"/>
-						<xs:attribute name="generatedBy" type="xs:anyURI" use="required"/>
-						<xs:attribute name="sourceId" type="xs:anyURI" use="optional"/>
-						<xs:attribute name="appliedDateTime" type="xs:dateTime"/>
-					</xs:complexType>
-				</xs:element>
+				<xs:element name="appliedProcessing" type="ebuttm:appliedProcessing_type" minOccurs="0"/>
```
nigelmegitt: This should have maxOccurs="unbounded" to match the spec.

ebu_tt_live/node/delay.py
```diff
@@ -42,7 +43,20 @@ def process_document(self, document, **kwargs):
...
-                # TODO: add an ebuttm:appliedProcessing element to the document metadata
+                if document.binding.head.metadata is None:
+                    document.binding.head.metadata = metadata.headMetadata_type(
+                        metadata.documentMetadata()
+                    )
+
+                if document.binding.head.metadata.documentMetadata is None:
+                    document.binding.head.metadata.documentMetadata = metadata.documentMetadata()
+
+                document.binding.head.metadata.documentMetadata.appliedProcessing = metadata.appliedProcessing_type(
```
nigelmegitt: What happens if there is already an appliedProcessing element? I did a test and found that it gets overwritten. In fact when the XSD is updated to allow any number of appliedProcessingElements this line fails - some other way to append the element is needed, which I have not yet worked out how to do...

nigelmegitt:
I have raised #438 to deal with the XSD issues etc; in the meantime this PR does add an `ebuttm:appliedProcessing` element when a retiming delay is applied.

ebu_tt_live/node/delay.py
nigelmegitt: If this is not needed it should be removed.

ebu_tt_live/node/delay.py
```diff
@@ -2,8 +2,9 @@
 import six
 import logging
 from .base import AbstractCombinedNode
-from datetime import timedelta
+from datetime import timedelta, datetime
```
SKHameed86: Is `from datetime import datetime` different? It won't import `timedelta` as well?

nigelmegitt: Not unless you tell it to - this is slightly confusing because the thing being imported and the module have the same name in the case of `datetime`.

ebu_tt_live/node/delay.py
tomrosier: Ah, I probably copy and pasted it from another bit of code.

ebu_tt_live/node/delay.py
tomrosier: Yep we need to import both :)

ebu_tt_live/node/delay.py
```diff
@@ -42,7 +43,24 @@ def process_document(self, document, **kwargs):
...
-                # TODO: add an ebuttm:appliedProcessing element to the document metadata
+                if document.binding.head.metadata is None:
+                    document.binding.head.metadata = metadata.headMetadata_type(
+                        metadata.documentMetadata()
+                    )
+
+                if document.binding.head.metadata.documentMetadata is None:
+                    document.binding.head.metadata.documentMetadata = metadata.documentMetadata()
+
+                if len(document.binding.head.metadata.documentMetadata.appliedProcessing) == 0:
```
nigelmegitt: It turns out you don't need to check for none existing already - I tried adding multiple and it worked! I'll push the change...

PR #437, opened 2017-05-23
415 deduplicator component

SKHameed86:


ebu_tt_live/bindings/__init__.py
```diff
@@ -39,7 +39,7 @@
 namespace_prefix_map = {
     'tt': raw.Namespace,
     'ebuttdt': ebuttdt.Namespace,
-    'ttp': ttp.Namespace,
+    'ttp': ttp.Namespace, , hn n nn
```
tomrosier: ` , hn n nn` ?

ebu_tt_live/bindings/__init__.py
```diff
@@ -654,6 +658,20 @@ def get_timing_type(self, timedelta_in):
...
 
+    def get_element_by_attr(self, attr, attr_type=None):
+        """
+        Lookup an attribute and return it. Optionally type is checked as well.
+        :param attr:
+        :param attr_type:
+        :return:
+        """
+        if self._attributes_by_name is None:
+            raise SemanticValidationError(ERR_SEMANTIC_VALIDATION_EXPECTED)
```
tomrosier: Style point: add a new line after if statements. Makes it easier to read

ebu_tt_live/bindings/__init__.py
```diff
@@ -825,6 +843,75 @@ def content_to_string(self, begin=None, end=None):
 
 raw.br_type._SetSupersedingClass(br_type)
 
+class style_attr(IDMixin, RegionedElementMixin, LiveStyledElementMixin,
+               SemanticValidationMixin, raw.source_style, raw.element_style)
+    def _source_style(self, dataset):
```
tomrosier: this and `__source_style_copy__` seem to be duplicates?

ebu_tt_live/bindings/__init__.py
```diff
@@ -825,6 +843,75 @@ def content_to_string(self, begin=None, end=None):
...
+        )
+        self._semantic_push_styles(dataset=dataset)
+
+    def _semantic_after_traversal(self, dataset, element_content=None, parent_binding=None):
+        self._semantic_postprocess_timing(dataset=dataset, element_content=element_content)
+
+    def _semantic_before_copy(self, dataset, element_content=None):
+        self._assert_in_segment(dataset=dataset, element_content=element_content)
+
+    def is_empty(self):
```
tomrosier: This seems overly written.

`return !(len(self.style) || len(self.span) || len(self.region) || len(self.div));` 

Typically multiple exit points are bad practice as they make the code difficult to read.

ebu_tt_live/bindings/__init__.py
```diff
@@ -825,6 +843,75 @@ def content_to_string(self, begin=None, end=None):
...
+
+        if len(self.region):
+            return False
+
+        if len(self.div):
+            return False
+
+        return True
+
+    def is_same(self):
```
tomrosier: don't really think this method is appropriate,

its a bit overly complicated

`return source_style.id == element_style.style`

is sufficent

ebu_tt_live/documents/ebutt3_deduplicator.py
```diff
@@ -0,0 +1,20 @@
+from datetime import timedelta
+import logging
```
tomrosier: Non selective imports are typically bad. Try and only import what you need.
In this case `from logging import getLogger` would be preferred.

ebu_tt_live/documents/test/deduplicator-blackbox.feature
```diff
@@ -0,0 +1,26 @@
+Feature: Deduplicator removes duplicated style and region elements
+
+  Examples:
+  | xml_file      |
+  | test_file.xml |
+
+  #Everything goes right
+  Scenario: The deduplicator successfully removes instances of element duplication from a file
```
tomrosier: there does not seem to be variables in your scenario that matches up with the examples below it?

ebu_tt_live/documents/test/deduplicator-unittests.feature
```diff
@@ -0,0 +1,45 @@
+Feature: Deduplicator removes duplicated elements
+
+  Background:
+    Given an xml file <xml_file>
+      | xml_file      |
+      | test_file.xml |
+
+  Scenario: assign new label to an element
+    Then the id is replaced by <new_label>
```
tomrosier: `id` should be `<xml_id>`

ebu_tt_live/documents/test/test_ebutt3deduplication.py
```diff
@@ -0,0 +1,198 @@
+from unittest import TestCase
```
tomrosier: Naming of file is actually not consistent with rest of repo it should be `test_ebutt_3_deduplication.py`

ebu_tt_live/bindings/__init__.py
SKHameed86: No idea what that was!

ebu_tt_live/bindings/__init__.py
SKHameed86: If I only need one should be fine to delete one, however I used the one for `div_type` as a template, and I'm thinking the difference is `source_style` grabs it from the document being tested, and `__source_style_copy__` makes sure that it's in the local instance.

At least, that's how I logicked it. Not sure if it's correct logic.

ebu_tt_live/bindings/__init__.py
SKHameed86: Is that suggestion to replace the code with?

tomrosier: yep :)

ebu_tt_live/bindings/__init__.py
SKHameed86: Done

ebu_tt_live/documents/ebutt3_deduplicator.py
SKHameed86: Done

ebu_tt_live/documents/test/deduplicator-unittests.feature
SKHameed86: Good spot!

ebu_tt_live/documents/test/test_ebutt3deduplication.py
SKHameed86: Done

ebu_tt_live/documents/test/deduplicator-blackbox.feature
SKHameed86: Think I've sorted it

ebu_tt_live/documents/ebutt3_deduplicator.py
```diff
@@ -1,20 +1,68 @@
 from datetime import timedelta
-import logging
+from logging import getLogger
 import copy
 from pyxb.binding.basis import NonElementContent, ElementContent
 from ebu_tt_live.bindings import tt
+from ..bindings.validation.base import SemanticValidationMixin
```
tomrosier: we want fixed paths rather than relative.

ebu_tt_live/documents/ebutt3_deduplicator.py
```diff
@@ -1,20 +1,68 @@
...
+
+    def __init__(self, style_id, style_attribute, sequence_identifier, sequence_number):
+        if style_id is None or style_attribute is None:
+            raise Exception()
+
+        self._style_id = list(style_id)
+        self._style_attribute = list(style_attribute)
+        self._sequence_idtentifier = sequence_identifier
+        self._sequence_number = sequence_number
+        self._perform_deduplication()
```
tomrosier: I get what your trying todo but i don't think this is the best approach. Usually the init is to just setup the class but not start execution.

ebu_tt_live/documents/ebutt3_deduplicator.py
```diff
@@ -1,20 +1,68 @@
...
+
+    class ValidateAttribute(RecursiveOperation):
```
tomrosier: class in a class, is not preferred. Makes things very difficult to test.

ebu_tt_live/documents/ebutt3_deduplicator.py
```diff
@@ -1,20 +1,68 @@
...
+        def __init__(self, root_element):
+            super(SemanticValidator, self).__init__(
+                root_element=root_element,
+                filter=self._attribute_validation_filter,
+                children_iterator='_validatedChildren'
+            )
+            self._attribute_dataset = {}
+
+        def _attribute_validation_filter(self, value, element):
+            if isinstance(value, SemanticValidationMixin):
```
tomrosier: `return isinstance(value, SemanticValidationMixin)` would be sufficient here

ebu_tt_live/documents/ebutt3_deduplicator.py
```diff
@@ -1,20 +1,68 @@
...
+
+        def __init__(self, root_element):
+            super(SemanticValidator, self).__init__(
+                root_element=root_element,
+                filter=self._attribute_validation_filter,
+                children_iterator='_validatedChildren'
+            )
+            self._attribute_dataset = {}
+
+        def _attribute_validation_filter(self, value, element):
```
tomrosier: `element` is unused?

ebu_tt_live/documents/ebutt3_deduplicator.py
```diff
@@ -1,20 +1,68 @@
...
+            if isinstance(value, SemanticValidationMixin):
+                return True
+            else:
+                return False
+
+        def _before_element(self, value, element=None, parent_binding=None, **kwargs):
+            value._semantic_before_traversal(dataset=self._attribute_dataset, element_content=element, parent_binding=parent_binding)
+            pass
```
tomrosier: is this a reminder to add more code here?

ebu_tt_live/documents/ebutt3_deduplicator-mock.py
```diff
@@ -0,0 +1,42 @@
...
+from pyxb import BIND
+
+style_1 = '<tt:style ebutts:linePadding="0.5c" tts:backgroundColor="rgb(0, 0, 0)" tts:color="rgb(255, 255, 255)" tts:fontFamily="sansSerif" xml:id="SEQ58.defaultStyle1"/>'
+style_2 = '<tt:style ebutts:linePadding="0.5c" tts:backgroundColor="rgb(0, 0, 0)" tts:color="rgb(255, 255, 255)" tts:fontFamily="sansSerif" xml:id="SEQ59.defaultStyle1"/>'
+old_style_list = []
+new_style_list = []
+style_1_list = []
+style_2_list = []
+
+line_padding_begin = style_1.find('linePadding="')
```
tomrosier: This seems very brittle. Maybe a regular expression would be better for finding these values.

nigelmegitt: Just so you're aware @tomrosier , @SKHameed86 and I looked at this on Friday and looked at how to do this properly with pyxb so we don't have to start text processing to fetch attributes at all.

ebu_tt_live/bindings/__init__.py
```diff
@@ -652,6 +656,20 @@ def get_timing_type(self, timedelta_in):
...
+        Lookup an attribute and return it. Optionally type is checked as well.
+        :param attr:
+        :param attr_type:
+        :return:
+        """
+        if self._attributes_by_name is None:
+            raise SemanticValidationError(ERR_SEMANTIC_VALIDATION_EXPECTED)
+        attribute = self._attributes_by_name.get(attr, None)
+        if attribute is None or attr_type is not None and not isinstance(attribute, attr_type):
+            raise LookupError(ERR_SEMANTIC_ELEMENT_BY_ID_MISSING.format(id=elem_id))
```
nigelmegitt: Seems like a strange error to throw when you've failed to find an attribute...

ebu_tt_live/node/deduplicator.py
```diff
@@ -53,43 +56,30 @@ def process_document(self, document, **kwargs):
...
-        new_style_list = [[],[],[],[]]
+            mirror_styles_no_id.add(mirror_entry_no_id)
 
-        i = 0
+            mirror_styles_new_id = list(mirror_styles_no_id)
 
-        list_a[0].append(original_styles[i].color)
-        list_a[1].append(original_styles[i].backgroundColor)
+            for x in enumerate(mirror_styles_new_id):
+                new_style = mirror_styles_new_id.append("xml:id=\"style\"" + str(x))
```
nigelmegitt: You shouldn't need to be putting the `xml:id=""` syntax in here - the values are extracted by pyxb and when you regenerate the XML pyxb will format it for you. You should only need to deal with the values themselves.

ebu_tt_live/node/deduplicator.py
```diff
@@ -53,43 +56,30 @@ def process_document(self, document, **kwargs):
...
 
-        for x in range(len(original_styles)):
-            list_b[0].append(original_styles[x].color)
-            list_b[1].append(original_styles[x].backgroundColor)
+                for span_style in enumerate(document.tt.body):
```
nigelmegitt: I don't think `document.tt.body` will contain _any_ span styles - if you want to do a hierarchical depth first search you'll need to change this code quite a bit to use recursive iteration or to use the RecursiveOperation class.

ebu_tt_live/node/deduplicator.py
```diff
@@ -53,43 +56,30 @@ def process_document(self, document, **kwargs):
...
+                    div.p.span.style = new_style
 
-            if list_a == list_b:
-                new_style_list[0].append(original_styles[i].id)
-                new_style_list[1].append(original_styles[i].color)
-                new_style_list[2].append(original_styles[i].backgroundColor)
-
-        for y in range(len(new_style_list[0])):
-            new_style_list[3].append("style" + str(y))
+                for new_styling in enumerate(mirror_styles_new_id):
```
nigelmegitt: You might want to remove the existing style children of `document.tt.head.styling` before adding the new ones...

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,85 @@
...
+
+log = logging.getLogger(__name__)
+document_logger = logging.getLogger('document_logger')
+
+
+class DeDuplicatorNode(AbstractCombinedNode):
+    _original_styles = []
+    _original_regions = []
+    _mirror_styles_no_id = []
+    _new_style_list = [[],[],[],[]]
```
tomrosier: 2d arrays, are not preferred. An array of style objects would be preferred.

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,85 @@
...
+    _new_region_list = [[],[],[],[]]
```
tomrosier: 2d arrays, are not preferred. An array of style objects would be preferred.

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,85 @@
...
+                self.limit_sequence_to_one(document)
+
+                # change the sequence identifier and sequence number
+                document.sequence_identifier = self._sequence_identifier
+                #document.sequence_number = self._sequence_number
+
+                document.validate()
+                self.producer_carriage.emit_data(data=document, **kwargs)
+
+    def remove_duplication(self, document, original_styles, original_regions, new_style_list, new_region_list, styling_element, region_element, span_style_id, region_style_id):
```
tomrosier: This is never called?

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,85 @@
...
+
+        for style in enumerate(document.tt.head.styling):
```
tomrosier: You should be able todo a object copy from `document.tt.head.styling` to `original_styles`

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,85 @@
...
+            original_styles.append(style)
+
+        # for region in enumerate(document.tt.head.layout):
+        #     original_regions.append(region)
+
+        mirror_styles = original_styles
```
tomrosier: I think this is just creating a reference rather than a copy.

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,85 @@
...
+
+            mirror_styles_no_id.add(mirror_entry_no_id)
+
+            mirror_styles_new_id = list(mirror_styles_no_id)
+
+            for x in enumerate(mirror_styles_new_id):
+                new_style = mirror_styles_new_id.append("xml:id=\"style\"" + str(x))
+
+                for span_style in enumerate(document.tt.body):
+                    div.p.span.style = new_style
```
tomrosier: `new_style` should be `span_style` shouldn't it?

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,85 @@
...
+
+                for new_styling in enumerate(mirror_styles_new_id):
```
tomrosier: could you not do `document.tt.head.styling` = `mirror_styles_new_id` iterating over the array seems a bit pointless?

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,199 @@
...
+            #deduplicating region elements
+            if value is not None:
+                unique_val = ComparableElement(value)
+                # stores references of original <xml:id> to <my_hash>
+                _old_id_dict[value.id] = unique_val.my_hash
+                # stores references of <my_hash> to <tt:region>
+                hash_dict[unique_val.my_hash] = value
+
+    def AppendNewElements(self, element_list, element_to_append_to, _old_id_dict, \
+                          _new_id_dict, hash_dict):
```
nigelmegitt: Seems a bit strange to put an underscore before function parameter names - you don't care about their scope with respect to a class here and they don't have to have the same name as the similar thing passed in.

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,199 @@
...
+        for hash_val in hash_dict:
+            new_id = hash_dict.get(hash_val)
```
nigelmegitt: Might be more elegant to change this loop a bit along the lines of:

```python
for hash_val, new_id in hash_dict.iteritems():
```

which does the lookup for you without making another call.

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,199 @@
...
+        return none_value
+
+
+class ComparableElement:
+    def __init__(self, value):
+        self.value = value
+
+        attributeDict = value._AttributeMap.copy()
+        xml_id_attr = ExpandedName('http://www.w3.org/XML/1998/namespace', 'id')
+        attributeDict.pop(xml_id_attr, None)
```
nigelmegitt: It's actually an error case if this is called with an element passed in that doesn't have an `xml:id` in this context - if that happens we should be raising an error somehow. We put the `None` in to silence that error, but it's troubling me that this might have happened, so it'd be good to dig a bit further.

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,199 @@
...
+
+        sortedDict = sorted(attributeDict.items(), key=lambda t: t[0])
```
nigelmegitt: Worth a comment to explain why we're sorting it, i.e. that we want to make sure that for two elements with the same set of attributes the values are put into the hash string in the same order.

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,199 @@
...
+        concatenatedStyleString = u''
+        for key,val in sortedDict:
+            styleValue = ReplaceNone(val.value(value))
+            concatenatedStyleString += str(styleValue)
+
+        for key,val in value.wildcardAttributeMap().items():
+            namespace = ReplaceNone(key.namespaceURI())
+            localName = key.localName()
+            wildcardValue = ReplaceNone(val)
+            concatenatedStyleString += namespace + localName + wildcardValue
```
nigelmegitt: Is there a test case that we can build to check that `nsa:foo="cat"` is distinguishable from `nsaf:oo="cat"` here? Could be that the concatenated namespace and localName coincide in this case.

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def _before_element(self, value, element=None, parent_binding=None, **kwargs):
+        pass
+
+    def _after_element(self, value, element=None, parent_binding=None, **kwargs):
+        pass
+
+    def _process_element(self, value, element=None, parent_binding=None, **kwargs):
+        # The latter part of this and the next test is to check that the instance
+        # is not a styling or layout element as these can also have style attributes
```
nigelmegitt: Slight tweak to the useful comment here: styling and layout can't have style attributes but their style _elements_ present themselves in exactly the same way as style attributes on other elements, so we have to avoid getting confused by them.

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,199 @@
...
+        if hasattr(value, 'style') and value.style is not None and not \
+        isinstance(value, bindings.styling):
+            id_to_index_dict = dict()
+
+            for old_id_index in range(len(value.style)-1, -1, -1):
```
nigelmegitt: Add comment about why we're going backwards.

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,199 @@
...
+                raise UnexpectedSequenceIdentifierError()
+
+            if self.check_if_document_seen(document=document):
+
+                self.limit_sequence_to_one(document)
+
+                document.sequence_identifier = self._sequence_identifier
+
+                self.remove_duplication(document=document)
+                print document.get_xml()
```
SKHameed86: Forgot I left that in!

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,199 @@
...
+                                                      _old_id_dict, \
+                                                      _new_id_dict)
+            replace_id_refs.proceed()
+        else:
+            return None
+
+    def CollateUniqueVals(self, element_list, _old_id_dict, _new_id_dict, \
+                          hash_dict):
+        for value in element_list:
+            #deduplicating in-line styles
```
SKHameed86: I've amended the comment.

From what I remember...we found that after the initial pass of deduplication, the deduplicator wasn't deduplicating in-line style attributes, so this is what this step is doing. I don't know if it needs to do the "step backwards" as in the `ReplaceStylesAndRegions` does it need to maintain the hierarchy?

ebu_tt_live/node/deduplicator.py
```diff
@@ -0,0 +1,199 @@
...
+            if value.style is not None:
+                for old_id_index in range(len(value.style)):
+                    old_id_ref = _old_id_dict.get(value.style[old_id_index])
+                    new_id_ref = _new_id_dict.get(old_id_ref)
+
+                    value.style[old_id_index] = new_id_ref
+
+            #deduplicating region elements
```
SKHameed86: I think there's a region element you've hurt the feelings of.

(I've amended the comment)

nigelmegitt: poor sensitive regions. They should know their place! geddit?!

ebu_tt_live/node/deduplicator.py
nigelmegitt: Well the same thing applies: if you have an attribute:

```xml
<style xml:id="s1" style="s2 s3 s4"/>
<style xml:id="s2" tts:color="yellow"/>
<style xml:id="s3" tts:color="green"/>
<style xml:id="s4" tts:color="yellow"/>
```

then `s2` and `s4` will be deduplicated and `s1` will end up with `style="s2 s3 s2"` which should be reduced to `style="s2 s3"` by removing the later occurrences of duplicates.

ebu_tt_live/node/deduplicator.py
SKHameed86: Removed the rogue underscore

ebu_tt_live/node/deduplicator.py
SKHameed86: Done

ebu_tt_live/node/deduplicator.py
SKHameed86: BDD Test done - @tomrosier pointed out something I had overlooked, but it passes. Giving it the old eyeball showed the correct style/element regions, but not sure if it passes in the right way.

ebu_tt_live/node/deduplicator.py
SKHameed86: Amended

ebu_tt_live/node/deduplicator.py
SKHameed86: Amended

docs/source/deduplication.rst
```diff
@@ -0,0 +1,40 @@
+DeDuplication of EBU-TT-Live documents
+======================================
+
+When documents are ReSequenced, duplication of ``style`` and ``region`` elements
+and attributes can occur. To address this, the DeDuplicator node takes in a
```
nigelmegitt: No need to put input carriage mech details in here - manifest file only applies for filesystem input. Presumably deduplicator can be configured with any input carriage mechanism...

ebu-debug.py
```diff
@@ -0,0 +1,4 @@
+import ebu_tt_live
```
tomrosier: This file shouldn't be included.

ebu_tt_live/bindings/__init__.py
```diff
@@ -825,6 +824,9 @@ def content_to_string(self, begin=None, end=None):
 raw.br_type._SetSupersedingClass(br_type)
 
 
+#raw.source_style._SetSupersedingClass(source_style)
```
tomrosier: Delete code artefacts.

docs/source/deduplication.rst
```diff
@@ -0,0 +1,40 @@
...
+``manifest_file``, and processes the document(s) with the
+:py:func:`ebu_tt_live.node.deduplicator.DeDuplicatorNode.remove_duplication` function.
+
+After copying ``styling`` and ``layout`` into a ``list()`` and setting them up for new
+``style`` and ``region`` elements, respectively, the ``list()`` is passed through the
+:py:func:`ebu_tt_live.node.deduplicator.DeDuplicatorNode.CollateUniqueVals` function.
+
+Because ``style`` and ``region`` elements can have ``style`` attributes, these
+are deduplicated first. Each element is then passed through the
```
nigelmegitt: Worth noting that a result of the deduplication might be two identical styles which differed only in the style attribute references before, but those references pointed at identical styles, so after, they end up looking the same. We don't do a repeat pass to deal with this, but that's a possible modification for later.

ebu_tt_live/documents/test/test_deduplication.py
```diff
@@ -0,0 +1,100 @@
+from unittest import TestCase
+import os
```
tomrosier: Is this used?

docs/source/deduplication.rst
```diff
@@ -0,0 +1,40 @@
...
+function, which takes in the list of elements, the path to the parent element
+(``styling`` or ``layout``), ``old_id_dict``, ``new_id_dict`` and ``hash_dict``
+
+The function iterates through the key-value pairs of ``hash_dict`` and the contents
+of the list of elements; where the ``xml:id`` of both matches, the element is appended to
+the parent element. The hash and ``xml:id`` is then stored in ``new_id_dict``,
+where the hash is the ``key`` and the ``xml:id`` is the ``value``.
+
+In the final step, before emitting the document, the document as it is at this stage is passed
+through the :py:class:`ebu_tt_live.node.deduplicator.ReplaceStylesAndRegions` class.
```
nigelmegitt: Worth saying this is a recursiveOperation, i.e. that the class is used to recursively fix the style and region attribute values.

docs/source/scripts_and_their_functions.rst
```diff
@@ -100,6 +100,14 @@ required. The Buffer Delay Node is primarily intended for delaying implicitly
 timed documents for resynchronisation. Use ``ebu-run`` to start this script, for
 example ``ebu-run --admin.conf=ebu_tt_live/examples/config/buffer_delay.json``
 
+DeDuplicator Node
+-----------------
+This node addresses the issues raised upon ReSequencing, where ``style`` and
```
nigelmegitt: Actually it will deduplicate any input, it's just that ReSequencing might cause such issues.

ebu_tt_live/documents/test/test_deduplicationComparableRegions.py
```diff
@@ -0,0 +1,101 @@
+from unittest import TestCase
```
tomrosier: This file seems identical to `ebu_tt_live/documents/test/test_deduplication.py`

ebu_tt_live/documents/test/test_deduplicationReplaceStylesAndRegions.py
```diff
@@ -0,0 +1,199 @@
+from unittest import TestCase
+import os
```
tomrosier: Is this used?

ebu_tt_live/documents/test/test_deduplicationReplaceStylesAndRegions.py
```diff
@@ -0,0 +1,199 @@
...
+            document.validate()
+
+            cdoc = self.replaceStylesAndRegions(document, self.test_old_ids, self.test_new_ids)
+            cdoc.proceed()
+
+            document.validate()
+
+            print document.get_xml()
+
+            # self.fail()
```
tomrosier: Remove commented code.

ebu_tt_live/bindings/__init__.py
```diff
@@ -825,6 +824,9 @@ def content_to_string(self, begin=None, end=None):
...
+#raw.element_style._SetSupersedingClass(element_style)
```
nigelmegitt: Not needed.

ebu_tt_live/documents/test/test_deduplicationComparableRegions.py
```diff
@@ -0,0 +1,101 @@
...
+        wrapOption='wrap',
+        padding='5px',
+        linePadding='0.5c'
+        )
+    comparableElement = deduplicator.ComparableElement
+
+    def test_dup_elements_diff_ids(self):
+        test_instance1 = self.comparableElement(self.test_style1)
+        test_instance2 = self.comparableElement(self.test_style2)
+        assert test_instance1.my_hash == test_instance2.my_hash
```
nigelmegitt: For all of these, should also check that the equality and non-equality operators work on the `comparableElement` objects. Are `assert_is_same` or `assert_different` available functions?

ebu_tt_live/documents/test/test_deduplicationReplaceStylesAndRegions.py
```diff
@@ -0,0 +1,199 @@
...
+
+            # self.test_new_ids.clear()
+            #
+            # check_attrs = CheckAttr(cdoc, self.test_new_ids)
+            # check_attrs.proceed()
+
+# class CheckAttr(RecursiveOperation):
```
nigelmegitt: Remove commented code

ebu_tt_live/documents/test/test_deduplication_tests.py
```diff
@@ -0,0 +1,92 @@
...
+            sequence_identifier = document_sequence,
+            producer_carriage = carriage
+        )
+
+    def test_process_document(self):
+        document = MagicMock(spec=EBUTT3Document)
+        self.deduplicatorNode.process_document(document)
+        self.deduplicatorNode.producer_carriage.emit_data.assert_called_with(data=document)
+
+    # def test_remove_duplication(self):
```
tomrosier: artefacts

ebu_tt_live/documents/test/test_deduplication_tests.py
```diff
@@ -0,0 +1,92 @@
...
+            color='rgb(255, 255, 255)'
+            )
+
+        test_styles = [test_style1, test_style2]
+        test_styles_blank = []
+
+        self.deduplicatorNode.CollateUniqueVals(test_styles, old_id_dict, new_id_dict, hash_dict)
+        self.deduplicatorNode.AppendNewElements(test_styles, test_styles_blank, old_id_dict, new_id_dict, hash_dict)
+
+        # assert len(test_styles_blank) is 1
```
tomrosier: artefact

testing/bdd/features/deduplicator/deduplicator-blackbox.feature
```diff
@@ -0,0 +1,77 @@
...
+    Then the second output document has <style_out_num_2> styles
+    And the second output document has <region_out_num_2> regions
+    And the second document is valid
+
+      Examples:
+        | xml_file_1                                 | style_out_num_1 | region_out_num_1 | xml_file_2                                 | style_out_num_2 | region_out_num_2 |
+        | deduplicator_templates/ReSequenced1_12.xml | 5               | 2                | deduplicator_templates/ReSequenced1_13.xml | 1               | 1                |
+
+
+#  Scenario: Upon receiving more than one file, the deduplicator reads each one sequentially and successfully removes instances of element duplication
```
tomrosier: artefact

ebu_tt_live/examples/config/producer_resequencer_consumer_fs.json
```diff
@@ -0,0 +1,46 @@
+{
```
nigelmegitt: Seems like for this PR we want this same thing with a deduplicator node between the resequencer and the consumer.

testing/bdd/test_deduplicator.py
```diff
@@ -0,0 +1,82 @@
...
+        producer_carriage=carriage
+    )
+    return instance
+
+
+@when('the document is processed')
+def when_document_processed(given_deduplicator_node, gen_document, test_context):
+    given_deduplicator_node.process_document(gen_document)
+    test_context['doc'] = gen_document
+#    given_deduplicator_node.producer_carriage.emit_data.assert_called_once()
```
tomrosier: artefact

testing/bdd/test_deduplicator.py
```diff
@@ -0,0 +1,82 @@
...
+    given_deduplicator_node.producer_carriage.reset_mock()
+
+@when('the first document is processed')
+def when_first_document_processed(given_deduplicator_node, gen_first_document, test_context):
+    given_deduplicator_node.process_document(gen_first_document)
+    test_context['doc1'] = gen_first_document
+#    given_deduplicator_node.producer_carriage.emit_data.assert_called_once()
```
tomrosier: artefact

testing/bdd/test_deduplicator.py
```diff
@@ -0,0 +1,82 @@
...
+    given_deduplicator_node.producer_carriage.reset_mock()
+
+@then('the second document is processed')
+def when_second_document_processed(given_deduplicator_node, gen_second_document, test_context):
+    given_deduplicator_node.process_document(gen_second_document)
+    test_context['doc2'] = gen_second_document
+#    given_deduplicator_node.producer_carriage.emit_data.assert_called_once()
```
tomrosier: artefact

testing/bdd/features/deduplicator/deduplicator-blackbox.feature
```diff
@@ -0,0 +1,77 @@
+Feature: Deduplicator removes duplicated style and region elements
```
nigelmegitt: We don't need to have "-blackbox" in the filename - it's a BDD so that's what it does.

testing/bdd/features/deduplicator/deduplicator-blackbox.feature
```diff
@@ -0,0 +1,77 @@
...
+      Examples:
+        | xml_file_1                                 | style_out_num_1 | region_out_num_1 | xml_file_2                                 | style_out_num_2 | region_out_num_2 |
+        | deduplicator_templates/ReSequenced1_12.xml | 5               | 2                | deduplicator_templates/ReSequenced1_13.xml | 1               | 1                |
+
+
+#  Scenario: Upon receiving more than one file, the deduplicator reads each one sequentially and successfully removes instances of element duplication
+#    Given the deduplicator receives more than one file
+
+  #Everything goes wrong
+#  Scenario: The deduplicator does nothing if no files are present
```
nigelmegitt: Remove commented lines

docs/source/deduplication.rst
SKHameed86: Removed that bit of the line

nigelmegitt: @SKHameed86 have you made any fixes here that you haven't yet pushed? I can't see any fix removing this bit...

ebu_tt_live/documents/test/test_deduplicationComparableRegions.py
SKHameed86: I can't find `assert_is_same` or `assert_different`, however, the test does check the equality and non-equality operators.

They're written as different scenarios, and if I were to add the other `testinstance`s, then all the scenarios would be the same.

nigelmegitt: The point here is that the tests only compare the _hashes_ and don't check that the `comparableElement` objects themselves have correctly functioning equality and non-equality operators.

Issue #440, opened 2017-05-24
Well timed children with badly timed siblings ignored


nigelmegitt:
Following the fix for #312, the computed begin and end times are incorrect when an element whose begin is later than the end has a sibling whose begin is before its end. Will create a branch demonstrating this and open a pull request...

nigelmegitt:
Inspiration for how to fix this: `RecursiveOperation` takes a `filter` - if we set that up for TimingValidationMixin to simply omit elements where begin is later than end, it should Just Work...

PR #441, opened 2017-05-24
Issue 440 siblings of badly timed elements

nigelmegitt:
Pull request for fixing issue #440.


ebu_tt_live/bindings/validation/timing.py
```diff
@@ -43,6 +43,13 @@ def _pre_init_variables(self, dataset, element_content):
         self._computed_end_time = None
         self._semantic_dataset = dataset
 
+    def _exclude_badly_timed_elements_filter(self, value, element):
```
tomrosier: Could this function be short handed?
`return True if not (element.begin is not None and element.end is not None and element.begin >= element.end) False` or something like that

ebu_tt_live/bindings/validation/timing.py
```diff
@@ -176,7 +183,10 @@ def _semantic_postprocess_timing(self, dataset, element_content):
             # This requires calculation based on the timings in its children.
 
             # All timing containers are complexTypes so we can call orderedContent safely
-            children = filter(lambda item: isinstance(item, TimingValidationMixin), [x.value for x in self.orderedContent()])
+            # but we don't want to bother with explicitly badly timed elements so filter
+            # them out.
+            children = filter(lambda item: isinstance(item, TimingValidationMixin) \
```
tomrosier: Not a fan with how long this line is.

testing/bdd/features/timing/computed_times.feature
```diff
@@ -11,6 +11,7 @@ Feature: Computed times computation
   # SPEC-CONFORMANCE: R16 R17 R132
   Scenario: Computed times of a document
     Given an xml file <xml_file>
+    And l is <l>
```
tomrosier: can this be more descriptive?

ebu_tt_live/bindings/validation/timing.py
nigelmegitt: I don't think Python 2.7 allows that. But I can change it to:

```python
return (element.begin is None or element.end is None or element.end > element.begin)
```

ebu_tt_live/bindings/validation/timing.py
nigelmegitt: Me neither. Looks like the only easy thing to do is to split the second argument to `filter(function, iterator)` onto a second line.

testing/bdd/features/timing/computed_times.feature
nigelmegitt: Yes, making it so...

ebu_tt_live/bindings/validation/timing.py
tomrosier: I think that would be the right thing todo :)

ebu_tt_live/bindings/validation/timing.py
nigelmegitt: OK, this was done though it is not showing as "outdated" on the GitHub pull request page strangely - @tomrosier if you're happy with it please could you "approve" in a review?

PR #445, opened 2017-06-02
Iss289 SMPTE

### LINKED ISSUE #289: Remove SMPTE-related code
SMPTE is dropped from v1.0 of 3350 so we do not need to deal with for now. No need for unspecced legacy so we should remove the code that involved with smpte.
This includes XSD
bindings, clocks

EyalLavi:
Partial fix for #289 since no code was removed.

### DIFF
--- a/docs/source/validation_framework.rst
+++ b/docs/source/validation_framework.rst
@@ -6,7 +6,7 @@ a pretty good job validating the syntax but it does not provide any tooling to p
 be described in XSD 1.0 (PyXB does not understand XSD 1.1 and xpath assertions).
 
 Example:
-    if tt:tt has timeBase=clock smpteTimingType must not be present anywhere in the document.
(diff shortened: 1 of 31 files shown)


docs/source/validation_framework.rst
```diff
@@ -6,7 +6,7 @@ a pretty good job validating the syntax but it does not provide any tooling to p
 be described in XSD 1.0 (PyXB does not understand XSD 1.1 and xpath assertions).
 
 Example:
-    if tt:tt has timeBase=clock smpteTimingType must not be present anywhere in the document.
+    If tt:timebase="clock" then all begin and end attributes of the subtitle shall be of type ebuttdt:clockTimingType.
```
nigelmegitt: needs to be `ttp:timebase`; would also change the wording of "the subtitle" to "timed elements".

PR #451, opened 2017-06-23
Serve ui pages

### LINKED ISSUE #450: UI components can't easily be served
We have a nice separation of shared styling assets into the `ui/assets` folder that can be used by `ui/test/test.html` and `ui/user_input_producer/user_input_producer.html` however this structure causes some problems:

1. The HTML pages are not called `index.html` so extra config in e.g. `httpd.conf` (for apache) is needed to serve those pages by default with a short URI
2. The links to the assets pages are relative (and reverse) links e.g. `../assets/css/...` which means that if the pages are served with an alias then those relative links are relative to the _alias_ URL, e.g. `http://machine/ebuttlive/ui/uip` as an alias would be resolved for the assets links to `http://machine/ebuttlive/assets/css/...` which typically would not exist. Of course it could also be added to as a new alias however that's just further config pain.

A better way to do this would be:

* [x] Call the index page for each web page `index.html`
* [x] Add to the makefile a build step that assembles the "servable" pages into a new location that can be served by GitHub Pages, copying in each required asset as a forward only relative link, e.g. to `css/main.css`
* [x] Modify the links in the HTML pages to point to these forward links.
* [x] Link to the pre-built versions of those pages from the LIT homepage and documentation

nigelmegitt:
Fixes #450.

* Makefile now cleans UI components from `build/docs/ui` and then makes the templates and copies all the components into place (`make ui`)
* Pages are all called `index.html` for default serving.
* Relative links to `../assets/*` replaced by forward links to `assets/*` - the contents of `assets` exists once in the repo and is duplicated into the relevant places by the Makefile
* Documentation updated to have fixed links and to link to the versions that will be served by GitHub Pages.

This breaks existing links to `ebu-tt-live/ui/user_input_producer/user-input-producer.html` in two ways:
1. The forward links to the assets no longer work since the `assets` folder is not in the place it is expecting - solution is to open the built versions instead.
2. The file no longer exists with that name - it is called `index.html` instead.

### DIFF
--- a/docs/source/scripts_and_their_functions.rst
+++ b/docs/source/scripts_and_their_functions.rst
@@ -22,7 +22,8 @@ into a sequence of EBU-TT Live documents using natural language processing. Use
 
 The default carriage mechanism is WebSocket, so you will need to listen to
 ``ws://127.0.0.1:9000``. Conveniently, we've created an HTML page that does just
-that. After you launch the Simple Producer, open ``docs/build/ui/test/index.html`` in your
+that. After you launch the Simple Producer, open ``docs/build/ui/test/index.html``
+`current release pre-built page <../ui/test>`_ in your
 browser. The 'Broadcast message' field should be populated with the correct
 address (``ws://localhost:9000``). Click 'Connect' and then 'Subscribe'. You can
 also change the identifier for the sequence. The documents should appear in the
@@ -56,7 +57,8 @@ Distributor or a Handover Manager Node. First, with your virtual environment
 activated and the code built, start one, with a command line such as  ``ebu-run
 --admin.conf ebu_tt_live/examples/config/user_input_producer_consumer.json`` -
 this one runs a simple consumer. Then, in your browser, open
-``docs/build/ui/user_input_producer/index.html`` and click
+``docs/build/ui/user_input_producer/index.html`` or the 
+`current release pre-built page <../ui/user_input_producer>`_ and click
 'Connect'. Select the sending mode (manual, scheduled or asynchronous). You
 should see the documents arriving in the command line window where the simple
 consumer is listening. See detailed instructions here:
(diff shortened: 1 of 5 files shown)


nigelmegitt:
Oh flip we have a build fail on `test_twisted_websocket`. Investigating...

nigelmegitt:
Raised build failure as #452

nigelmegitt:
OK now we build, if anyone would like to review that'd be great - @EyalLavi ?

docs/source/scripts_and_their_functions.rst
```diff
@@ -23,7 +23,7 @@ into a sequence of EBU-TT Live documents using natural language processing. Use
 The default carriage mechanism is WebSocket, so you will need to listen to
 ``ws://127.0.0.1:9000``. Conveniently, we've created an HTML page that does just
 that. After you launch the Simple Producer, open ``docs/build/ui/test/index.html``
-or the `current release pre-built page <../ui/test>`_ in your
+or the `current release pre-built page <../ui/test/index.html>`_ in your
```
nigelmegitt: as just discussed, the link doesn't go to the current release pre-built page anyway, since it's a relative link - it will just go to the current built version wherever that is. Probably should add this hyperlink around the link text earlier in the sentence, and change this hyperlink to be hard coded to the ebu.github.io one.

docs/source/scripts_and_their_functions.rst
```diff
@@ -57,8 +57,8 @@ Distributor or a Handover Manager Node. First, with your virtual environment
 activated and the code built, start one, with a command line such as  ``ebu-run
 --admin.conf ebu_tt_live/examples/config/user_input_producer_consumer.json`` -
 this one runs a simple consumer. Then, in your browser, open
-``docs/build/ui/user_input_producer/index.html`` or the 
-`current release pre-built page <../ui/user_input_producer>`_ and click
+``docs/build/ui/user_input_producer/index.html`` or the
+`current release pre-built page <../ui/user_input_producer/index.html>`_ and click
```
nigelmegitt: Same comment as above.

Issue #452, opened 2017-06-23
Twisted WebSocket test build failure


nigelmegitt:
Unexpected build failure with no change in our code:

```Shell
================================================================================== FAILURES ==================================================================================
______________________________________________________ TestProdServerToConsClientProtocols.test_url_encoded_components _______________________________________________________

self = <test_twisted_websocket.TestProdServerToConsClientProtocols testMethod=test_url_encoded_components>

    def test_url_encoded_components(self):
        # This test is about getting percent encoded characters work in sequenceId or hostname
        sequence_id = u'sequence/ünicödé?/Name'
        self._create_server(url='ws://localhost:9006', producer=self.prod)
        self._create_client(
            url='ws://localhost:9006/sequence%2F%C3%BCnic%C3%B6d%C3%A9%3F%2FName/subscribe',
            consumer=self.cons
        )
    
        self.cproto.consumer = self.cons
    
>       self._connect()

ebu_tt_live/twisted/test/test_twisted_websocket.py:214: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ebu_tt_live/twisted/test/test_twisted_websocket.py:47: in _connect
    self.assertEquals(self.sproto.state, self.sproto.STATE_OPEN)
venv/lib/python2.7/site-packages/twisted/trial/_synctest.py:432: in assertEqual
    super(_Assertions, self).assertEqual(first, second, msg)
E   FailTest: 0 != 3
-------------------------------------------------------------------------------- Captured log --------------------------------------------------------------------------------
websocket.py               227 ERROR    one or more reserved delimiters /?# present in path segment: u'sequence/\xfcnic\xf6d\xe9?/Name'
websocket.py               352 ERROR    one or more reserved delimiters /?# present in path segment: u'sequence/\xfcnic\xf6d\xe9?/Name'
```

Presumably this has been induced by a change to a dependency.

nigelmegitt:
[...]

PR #455, opened 2017-07-07
Issue 454 encoder timebase on emit

### LINKED ISSUE #454: EBU-TT-D encoder with filesystem output carriage mech discards all docs
Using this config file:

```json
{
  "nodes": {
    "node1": {
      "id": "producer1",
      "type": "simple-producer",
      "show_time": true,
      "sequence_identifier": "TestSequence1",
      "output": {
        "carriage": {
          "type": "direct",
          "id": "pipe1"
        }
      }
    },
    "node2": {
      "id": "resequencer1",
      "type": "resequencer",
      "sequence_identifier": "ReSequenced1",
      "segment_length": "5.0",
      "input": {
        "carriage": {
          "type": "direct",
          "id": "pipe1"
        }
      },
      "output": {
        "carriage": {
          "type": "direct",
          "id": "pipe2"
        }
      }
    },
    "node3": {
      "id": "encoder1",
      "type": "ebuttd-encoder",
      "input": {
        "carriage": {
          "type": "direct",
          "id": "pipe2"
        }
      },
      "output": {
        "carriage": {
          "type": "filesystem"
        }
      }
    }
  }
}
```

produces warnings such as:

```
[WARNING] (2017-07-07 17:10:02,050) in ebu_tt_live.carriage.filesystem[131] - Could not resolve availability time for item in sequence: default. Item discarded.
```

every time a document is due to be emitted, and therefore it emits no documents.

On digging, I see that this is because `EBUTTDEncoder.process_document()` calls `emit_data()` without specifying a `time_base` parameter, and `FilesystemProducerImpl.emit_data()` needs it to get the relevant clock.

nigelmegitt:
Fixes #454 by specifying the `time_base` parameter when emitting from the Encoder node. `FilesystemProducerImpl.emit_data()` is then able to generate a suitable clock for the media timebase.

Hard coded to media timebase because that’s the only kind that’s permitted for EBU-TT-D documents. Arguably we should pass through whatever the document says, but I decided not to due to the pointless overhead. One day this will trip me up, possibly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,8 +109,9 @@ gen
 # Generated bindings should not appear in the repo
 ebu_tt_live/bindings/raw
 
-# Test export files from file system carriage
+# Test and non-test export files from file system carriage
 test_export/
+export/
 
 # Node.js file
 node_modules/
@@ -127,3 +128,4 @@ docs/build/*
 
 # Mac OS X files
 .DS_Store
+
--- a/ebu_tt_live/examples/config/sproducer_resequencer_direct_ebuttd_encoder_fs.json
+++ b/ebu_tt_live/examples/config/sproducer_resequencer_direct_ebuttd_encoder_fs.json
@@ -1,0 +1,49 @@
+{
+  "nodes": {
+    "node1": {
+      "id": "producer1",
+      "type": "simple-producer",
+      "show_time": true,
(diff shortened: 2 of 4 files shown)


nigelmegitt:
The build failure is due to #452 only causing the WebSocket URI encoding test to fail - all other tests pass.

PR #462, opened 2017-09-15
Issue 461 xsd fix

### LINKED ISSUE #461: authorsGroupIdentifier wrongly specified in XSD
`ebutt_parameters.xsd` defines the `authorsGroupIdentifier` attribute with both a `type` attribute and a `simpleType` element. Oxygen complains that this is an error, and the `type` attribute should be removed.

nigelmegitt:
Fix #461 by removing superfluous and erroneous type attribute.

### DIFF
--- a/ebu_tt_live/documents/test/converter_ericsson1.xml
+++ b/ebu_tt_live/documents/test/converter_ericsson1.xml
@@ -12,7 +12,7 @@
                 <ebuttm:documentOriginalProgrammeTitle>tgv</ebuttm:documentOriginalProgrammeTitle>
                 <ebuttm:documentOriginalEpisodeTitle>tgv</ebuttm:documentOriginalEpisodeTitle>
                 <ebuttm:documentCreationDate>2016-09-06</ebuttm:documentCreationDate>
-                <!--<ebuttm:documentRevisionNumber/> <== this element was part of the document originally generated but fails the tests because it's not conformant, so commented out in order to proceed, but leaving in for future archaeologists -->
+                <!--<ebuttm:documentRevisionNumber/> --> <!-- <== this element was part of the document originally generated but fails the tests because it's not conformant, so commented out in order to proceed, but leaving in for future archaeologists -->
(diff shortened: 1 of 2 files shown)


ebu_tt_live/documents/test/converter_ericsson1.xml
```diff
@@ -12,7 +12,7 @@
                 <ebuttm:documentOriginalProgrammeTitle>tgv</ebuttm:documentOriginalProgrammeTitle>
                 <ebuttm:documentOriginalEpisodeTitle>tgv</ebuttm:documentOriginalEpisodeTitle>
                 <ebuttm:documentCreationDate>2016-09-06</ebuttm:documentCreationDate>
-                <ebuttm:documentRevisionNumber/>
+                <!--<ebuttm:documentRevisionNumber/>-->
```
tomrosier: Don't like the artefact.

nigelmegitt: Yeah, it's not nice but since it was an Ericsson generated example it felt bad to remove it completely, which would have hidden the fact that their tool generated it. On balance I preferred to keep the artefact then remove it altogether. I'm going to add a note into the comment explaining this better.

ebu_tt_live/documents/test/converter_ericsson1.xml
```diff
@@ -12,7 +12,7 @@
                 <ebuttm:documentOriginalProgrammeTitle>tgv</ebuttm:documentOriginalProgrammeTitle>
                 <ebuttm:documentOriginalEpisodeTitle>tgv</ebuttm:documentOriginalEpisodeTitle>
                 <ebuttm:documentCreationDate>2016-09-06</ebuttm:documentCreationDate>
-                <!--<ebuttm:documentRevisionNumber/>-->
+                <!--<ebuttm:documentRevisionNumber/> <== this element was part of the document originally generated but fails the tests because it's not conformant, so commented out in order to proceed, but leaving in for future archaeologists -->
```
tomrosier: <== ? shouldn't it be <!-- ?

Issue #464, opened 2017-09-22
Add deduplicator node to EBU-TT-D encoding config


nigelmegitt:
In other words, make a config file that tidily makes EBU-TT-D.

nigelmegitt:
This is done in v2.1.1 release.

PR #465, opened 2017-09-22
Issue 464 deduplicator before ebuttd encoding

nigelmegitt:
Fix for #464. Should deduplicate resequenced output before EBU-TT-D
encoding, and make tidy EBU-TT-D.


nigelmegitt:
This has no code changes, only config file changes, and works fine on my machine, so I plan to go ahead and merge without a review when the build has passed. We probably need to do a general check and tidy-up of all the config files, but I'll raise that as a separate issue.

PR #470, opened 2018-02-22
Replace (broken) slack invitation link with e-mail

frans-ebu:


frans-ebu:
Ok, I have created #471

PR #471, opened 2018-02-23
Replace (broken) Slack invite link with e-mail

frans-ebu:


nigelmegitt:
Changed base to 2.1.1, so we can make this part of a minor point release. May as well do it "properly"...

nigelmegitt:
I'm not sure why it isn't building though, seems like there's something wrong with the openssl command, maybe the encryption keys need to be refreshed?

frans-ebu:
@nigelmegitt I don't think anything is wrong. This is just a protection against security risks from PRs originating from a forked repository (mine in this case).

See: https://docs.travis-ci.com/user/pull-requests#Pull-Requests-and-Security-Restrictions

So options are:
- Merge anyway
- Create a PR in this git directly

nigelmegitt:
Oh neat! I didn't spot that you did it from a fork. No reason I can think of why not to make the PR in this repository directly. That's what I would have expected to be honest.

frans-ebu:
Does not deploy, as PR comes from a fork (encryption keys, security measure).

PR #473, opened 2018-02-23
Release/2.1.1

nigelmegitt:
Minor point release? Quite a lot of fixes:

* Adds deduplicator
* Fixes a build destination problem
* Removes `filesystem-simple` carriage mechanism, moving its functionality into `filesystem`
* Filter out badly timed elements as per the spec
* Include non-maintenance comment in direct scripts (in favour of `ebu-run`)
* Fix hyperlink dependency issue (temporary fix hopefully)
* Change documentation build and URLs in documentation so they don't use backwards relative links
* Fix XSD to allow foreign namespace content as per spec
* Add line numbers to computed line BDD tests so that failures can be located
* Fix documentation home page slack sign-up link to be an email request, since the linked page no longer works.


nigelmegitt:
I'm slightly baffled that a comment I made back in June, associated with a commit, has magically appeared here. The comment is correct though, I guess the pull request was merged without it being addressed, and I'm about to push that on for the sake of progress :-(

frans-ebu:
Great to see this being released with all tests passed! Many tnx @nigelmegitt and @SKHameed86

PR #475, opened 2018-02-23
User relative links to templates

### LINKED ISSUE #474: Served UIP doesn't send live handover messages
probably because the relative links are relative to the wrong thing

nigelmegitt:
Should fix #474.

Uses relative links to templates. (how did it work at all before?)
Put message payload into the UIP display so that the receiver can read it without having to open the message XML up.


nigelmegitt:
This looks fine to merge.

nigelmegitt:
Merging now ahead of future work to move to Python3.

PR #482, opened 2018-05-08
Issue 481 counter number

### LINKED ISSUE #481: [EBUTTDEncoder] Allow first output file number to be specified
The `FilesystemProducerImpl` always begins output file numbers at `1` when it is first run, for EBU-TT-D documents. Allow it to be specified as a config parameter.

Rationale:

When creating EBU-TT-D documents to be wrapped in MP4 and served with a DASH manifest, the process may need to be begun some time after the MPD's `availabilityStartTime`. When this happens, in a live mode, the segment number needs to be the `(currentTime - availabilityStartTime)/segmentDuration` - that could be calculated in the encoder, or alternatively it could be preset in a config file that is modified to set the right value.

nigelmegitt:
Fixes #481 by allowing the counter value to be specified.

### DIFF
--- a/docs/source/configurator.rst
+++ b/docs/source/configurator.rst
@@ -106,8 +106,11 @@ Node type dependent options for [nodeN] : ::
    type="ebuttd-encoder"
    ├─media_time_zero : ["current" (default) | clock time at media time zero TODO: check format]
    ├─default_namespace : ["false" (default) | "true"]
-   └─clock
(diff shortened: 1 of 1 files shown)


nigelmegitt:
I've just re-reviewed this.
* No unit tests => can't merge at this time.

This seems to bring in some changes also present in other pull requests, I'm not sure if there's an ordering dependency on merging them.

PR #486, opened 2018-05-09
Issue 485 remote clock

### LINKED ISSUE #485: [EBUTTDENCODER] Fetch a clock value from a remote URL
Use a remote server to get the current time.

nigelmegitt:
Fix #485 by allowing a remote URL to be provided, which is dereferenced and the returned ISO date processed and used.


nigelmegitt:
I've just re-reviewed this. Some thoughts:
* No unit tests on this code => not merge-able yet.
* Otherwise it is somewhat useful for resequencing at known times, but does not in itself ensure that the extracted segments timings are locked to the intended times. All it does is makes the requests for segments at correct times: it does not specify the segment times being requested to match. That in turn creates a buggy behaviour where there are edge effects around the beginning and end of each segment extracted, when converting to EBU-TT-D and serving as live subtitles in fragmented MP4.

Conclusion: more work needed.

PR #489, opened 2019-07-17
Python3 migration

### LINKED ISSUE #488: Move to Python 3
Python 2.7 is going out of support, so we need to move to Python 3.

Issues currently:
* `ebu_tt_live/utils.py", line 117` uses `async` as a function parameter label, but in python 3 it is a reserved keyword. Change the label.
* `ebu_tt_live/utils.py", line 4` uses `Queue` but it should be `queue` - python 3 is case sensitive.
* pyxb generates `ebu_tt_live/bindings`but the `_ttm` module cannot be imported.

PyvesB:
This pull request closes #488.

Each fix was broken down in a separate commit with a (hopefully) informative message, however, we could squash all commits when merging to master.


nigelmegitt:
Thank you @PyvesB for your great work on this. 

I see that there's an issue with travis CI to do with encrypted environment variables so I will check it manually. I also expect a BDD failure matching #490 to occur, since I haven't found the fix for that yet.

One other thing, I will create a release candidate branch and make that the target of this pull request rather than heading directly to master.

nigelmegitt:
I tested this manually and get a bunch of test failures, all in `ebu_tt_live/twisted/test/test_twisted_websocket.py` except for one that relates to #490. @PyvesB I think perhaps you've resolved this somehow?

PyvesB:
These failures aren't dependent on Python versions, but rather the version of the Autobahn dependency. The 1.18.x and 1.19.x versions no longer allow the test suite you mentioned to run successfully.

nigelmegitt:
Thanks @PyvesB then I'll push a change to `setup.py` to restrict the autobahn version to <18 while we think about how to fix those tests.

nigelmegitt:
I note that this pull request has a problem that other pull requests from forks have, that the travis CI build cannot run because the source repo doesn't have the same credentials as the destination repo. I'm not sure what pattern fixes this, but the easiest practical thing is to merge this into a new branch in the ebu repo and pull request from that into `release/3.0`.

nigelmegitt:
Closing in favour of #512 which should have more success with the travis CI build.

Issue #502, opened 2019-10-02
Resequencer: document w/o content always has sequence number 1


spoeschel:
When the Resequencer is used, it may happen that in the meantime no new Part 3 segments are queued and hence the regular output document is a document without content (which is a special case in the code). However this empty document always has a `ebuttp:sequenceNumber` of `1` instead of the actual applicable sequence number. In case segment/content is available, in contrast a correct sequence number is used and the requirement of a strictly monotonic increasing sequence number is satisfied.

In the end the reason is that the [EBUTT3Splicer](https://github.com/ebu/ebu-tt-live-toolkit/blob/master/ebu_tt_live/documents/ebutt3.py#L1003) is provided with the current sequence number the [Resequencer](https://github.com/ebu/ebu-tt-live-toolkit/blob/master/ebu_tt_live/node/consumer.py#L68) stores (and increments, when an output document is issued) - while [create_compatible_document](https://github.com/ebu/ebu-tt-live-toolkit/blob/master/ebu_tt_live/documents/ebutt3.py#L997) relies on the sequence number stored internally by [EBUTT3DocumentSequence](https://github.com/ebu/ebu-tt-live-toolkit/blob/master/ebu_tt_live/documents/ebutt3.py#L572).

nigelmegitt:
I've been looking at #509 and the comments on it and I'm now somewhat confused about what the issue actually is.

The current behaviour is:
* Until any documents have been received, then call `create_compatible_document` which always makes an empty document with a sequence number of 1. Since this is the same document every time, and the receiving node can discard duplicates, this causes no problem.
* As soon as any documents have been received, `not document_segments` evaluates to `False` and  `EBUTT3Splicer` is called, which uses an incrementing sequence number. Even if the splicer extracts an empty document, it should still use a correctly incrementing sequence number.

Also, in relation to:

> the requirement of a strictly monotonic increasing sequence number

There is no such requirement. Sequence numbers have to increase, but not necessarily monotonically.

So now I'm wondering what test scenario results in unexpected empty documents with sequence number 1, when a different number should be used? @spoeschel please could you explain a bit more about what behaviour you're seeing and exactly what the problem is?

spoeschel:
> * Until any documents have been received, then call `create_compatible_document` which always makes an empty document with a sequence number of 1. Since this is the same document every time, and the receiving node can discard duplicates, this causes no problem.

Agreed.

[...]

PR #503, opened 2019-10-04
Resequencer: fix sequence number 1 if no content

spoeschel:
While the EBUTT3Splicer is provided with the current sequence number
stored/incremented by the Resequencer, the `create_compatible_document`
method so far refers to the sequence number stored internally by
EBUTT3DocumentSequence instead. This leads to a sequence number of 1 in
the issued output document if no subtitle content is available.

To align with the correct (strictly monotonic increasing) sequence
number, always use the one stored by the Resequencer.

Fixes ebu/ebu-tt-live-toolkit#502.


nigelmegitt:
Hi @spoeschel please could you rebase this on the `release/2.1.2` branch? Our release planning pattern here is to open pull requests against a release stage branch, then when we have a good candidate for release, merge that wholesale into `master` and add a release tag.

spoeschel:
Ah, I see; this makes sense. I rebased the MR branch and (force) pushed it again.

nigelmegitt:
Thank you! @frans-ebu is on the case to sort out the travis build issue.

spoeschel:
Ah, it seems that I have to make a MR from a branch on this very repo (instead of my fork's branch), as otherwise [the encrypted variables](https://docs.travis-ci.com/user/pull-requests#pull-requests-and-security-restrictions) needed by the build are not available.

So I will close this MR and #506 then, and open new MRs (as I cannot edit the source repo/branch here).

nigelmegitt:
You now have write permissions on the repo @spoeschel so you should be able to do that. Thank you for your patience!

Issue #505, opened 2019-10-07
Resequencer: allow to immediately issue documents


spoeschel:
So far the resequencer starts to regularly issue documents only after the first EBU-TT Live document has been received. However for some use cases this might be inconvenient as an active document must exist at all times e.g. when creating segments for an MPEG-DASH stream. The reason for documents being issued only after the first received EBU-TT Live document is that certain parameters of that document are used for initialisation.

nigelmegitt:
Yes, this is true. I guess perhaps some kind of init document could be processed to initialise the resequencer.

spoeschel:
I think this is the most elegant and future-proof way to address this; I will prepare a MR...

nigelmegitt:
Many thanks!

PR #506, opened 2019-10-07
Resequencer: allow to immediately issue documents

spoeschel:
So far the resequencer starts to regularly issue documents only after
the first EBU-TT Live document has been received. However for some use
cases this might be inconvenient as an active document must exist at all
times e.g. when creating segments for an MPEG-DASH stream. The reason
for documents being issued only after the first received EBU-TT Live
document is that certain parameters of that document are used for
initialisation.

This commit adds a new configuration parameter that specifies a document
which will be used for initialisation (instead of the first received
EBU-TT Live document). Therefore the resequencer will immediately (be
able to) start issuing documents after its creation.

Closes ebu/ebu-tt-live-toolkit#505.

### DIFF
--- a/docs/source/configurator.rst
+++ b/docs/source/configurator.rst
@@ -99,6 +99,7 @@ Node type dependent options for [nodeN] : ::
    ├─sequence_identifier : sequence identifier, default "re-sequencer"
    ├─segment_length : duration of each output segment in seconds, default 2
    ├─begin_output : ["immediate" (default) | {begin time} ] the time at which the first output segment should begin.
+   ├─init_document : File path to a document used to set initial parameters and afterwards start issuing documents (instead of awaiting the first received document), default None
    ├─discard : whether to discard content that has been encoded, default True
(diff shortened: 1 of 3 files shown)


nigelmegitt:
Hi @spoeschel please could you rebase this on the `release/2.1.2` branch? Our release planning pattern here is to open pull requests against a release stage branch, then when we have a good candidate for release, merge that wholesale into `master` and add a release tag.

spoeschel:
Yes, I rebased the MR branch and (force) pushed it again.

nigelmegitt:
Thank you! @frans-ebu is on the case to sort out the travis build issue.

nigelmegitt:
@spoeschel the travis build issue has now been fixed but I think it may need yet another rebase, apologies.

spoeschel:
Closing, because of https://github.com/ebu/ebu-tt-live-toolkit/pull/503#issuecomment-540489252.

PR #508, opened 2019-10-09
Issue 507 timedelta

### LINKED ISSUE #507: as_timedelta incorrectly interprets milliseconds
If a `FullClockTimingType` or `LimitedClockTimingType` has a decimal fraction of seconds component that is not 3 digits long the milliseconds value is incorrectly assigned.

For example: `00:00:00.17` gets interpreted as being 17 milliseconds instead of 170 milliseconds.

nigelmegitt:
Fixes #507 by interpreting decimal fractions of seconds as milliseconds, up to 3 decimal places, and adding tests to verify this behaviour.

### DIFF
--- a/ebu_tt_live/bindings/_ebuttdt.py
+++ b/ebu_tt_live/bindings/_ebuttdt.py
@@ -38,13 +38,6 @@ class _TimedeltaBindingMixin(object):
     }
 
     @classmethod
-    def _int_or_none(cls, value):
-        try:
-            return int(value)
-        except TypeError:
-            return 0
-
-    @classmethod
     def compatible_timebases(cls):
         return cls._compatible_timebases
 
@@ -320,10 +313,10 @@ class FullClockTimingType(SemanticValidationMixin, _TimedeltaBindingMixin, ebutt
         :return:
         """
         hours_str, minutes_str, seconds_str, seconds_fraction_str = [x for x in cls._groups_regex.match(instance).groups()]
-        milliseconds = seconds_fraction_str and cls._int_or_none('{:0<3}'.format(seconds_fraction_str)[:3]) or 0
-        return timedelta(hours=cls._int_or_none(hours_str), 
-            minutes=cls._int_or_none(minutes_str), 
-            seconds=cls._int_or_none(seconds_str), 
+        milliseconds = seconds_fraction_str and float('0.' + seconds_fraction_str) * 1000 or 0
+        return timedelta(hours=int(hours_str), 
+            minutes=int(minutes_str), 
+            seconds=int(seconds_str), 
             milliseconds=milliseconds)
 
     @classmethod
@@ -372,10 +365,10 @@ class LimitedClockTimingType(_TimedeltaBindingMixin, ebuttdt_raw.limitedClockTim
         :return:
         """
         hours_str, minutes_str, seconds_str, seconds_fraction_str = [x for x in cls._groups_regex.match(instance).groups()]
-        milliseconds = seconds_fraction_str and cls._int_or_none('{:0<3}'.format(seconds_fraction_str)[:3]) or 0
-        return timedelta(hours=cls._int_or_none(hours_str), 
-            minutes=cls._int_or_none(minutes_str), 
-            seconds=cls._int_or_none(seconds_str), 
+        milliseconds = seconds_fraction_str and float('0.' + seconds_fraction_str) * 1000 or 0
+        return timedelta(hours=int(hours_str), 
+            minutes=int(minutes_str), 
+            seconds=int(seconds_str), 
             milliseconds=milliseconds)
         
 
--- a/ebu_tt_live/bindings/_ebuttdt.py
+++ b/ebu_tt_live/bindings/_ebuttdt.py
@@ -319,11 +319,12 @@ class FullClockTimingType(SemanticValidationMixin, _TimedeltaBindingMixin, ebutt
         :param instance:
         :return:
         """
-        hours, minutes, seconds, milliseconds = map(
-            lambda x: cls._int_or_none(x),
-            cls._groups_regex.match(instance).groups()
-        )
-        return timedelta(hours=hours, minutes=minutes, seconds=seconds, milliseconds=milliseconds)
+        hours_str, minutes_str, seconds_str, seconds_fraction_str = [x for x in cls._groups_regex.match(instance).groups()]
+        milliseconds = seconds_fraction_str and cls._int_or_none('{:0<3}'.format(seconds_fraction_str)[:3]) or 0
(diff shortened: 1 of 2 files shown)


ebu_tt_live/bindings/test/test_times.py
```diff
@@ -0,0 +1,72 @@
...
+        '1m': timedelta(minutes = 1),
+        '1.3h': timedelta(hours = 1, minutes = 18),
+        '1.03h': timedelta(hours = 1, minutes = 1, seconds = 48),
+        '1.30h': timedelta(hours = 1, minutes = 18),
+        '1h': timedelta(hours = 1),
+        '1.3ms': timedelta(milliseconds = 1, microseconds = 300),
+        '1.03ms': timedelta(milliseconds = 1, microseconds = 30),
+        '1.30ms': timedelta(milliseconds = 1, microseconds = 300),
+        '1ms': timedelta(milliseconds = 1),
+    }
```
spoeschel: As up to three decimal places are supported, the tests should check the (corner) case of three decimal places, too.

ebu_tt_live/bindings/_ebuttdt.py
```diff
@@ -319,11 +319,12 @@ def as_timedelta(cls, instance):
         :param instance:
         :return:
         """
-        hours, minutes, seconds, milliseconds = map(
-            lambda x: cls._int_or_none(x),
-            cls._groups_regex.match(instance).groups()
-        )
-        return timedelta(hours=hours, minutes=minutes, seconds=seconds, milliseconds=milliseconds)
+        hours_str, minutes_str, seconds_str, seconds_fraction_str = [x for x in cls._groups_regex.match(instance).groups()]
+        milliseconds = seconds_fraction_str and cls._int_or_none('{:0<3}'.format(seconds_fraction_str)[:3]) or 0
```
spoeschel: Decimal places beyond the third one are truncated here, so e.g. a timecode of `12:34:56.399999` is interpreted as `12:34:56.399`. It could make sense to instead refer to the fourth decimal place for proper rounding (leading to `12:34:56.400` here), in order to fix timestamps of imprecise handling upstream. So maybe something like `int(round(float('0.' + seconds_fraction_str), 3) * 1000)`.

ebu_tt_live/bindings/_ebuttdt.py
```diff
@@ -319,11 +319,12 @@ def as_timedelta(cls, instance):
...
+        return timedelta(hours=cls._int_or_none(hours_str),
```
spoeschel: Hmm, is there a specific reason why `_int_or_none` is called here (and at other locations)? The regex should already ensure that the affected value is a valid integer, shouldn't it?

ebu_tt_live/bindings/test/test_times.py
spoeschel: And maybe also the case of four decimal places...

ebu_tt_live/bindings/_ebuttdt.py
nigelmegitt: No, it doesn't. If the group is missing then it returns None and then we get an exception downstream. In particular the last group can be missing.

ebu_tt_live/bindings/_ebuttdt.py
nigelmegitt: Interesting approach, I'll try that.

ebu_tt_live/bindings/test/test_times.py
nigelmegitt: I've added a fix for up to 3 decimal places.

ebu_tt_live/bindings/_ebuttdt.py
spoeschel: If the regex doesn't match, the `match` call [already returns `None` instead of a MatchObject](https://docs.python.org/2.7/library/re.html#re.RegexObject.match) (which would provide the `groups` method). So an exception is thrown in line 322 then (tested with value `10:00X:00.00`):

    AttributeError: 'NoneType' object has no attribute 'groups'

Consequently the invokations of `_int_or_none` have no effect here and could be removed.

spoeschel: (except in case of the last group I think)

nigelmegitt: Did you test with `10:00:00`?

spoeschel: With `10:00:00` it indeed works with `_int_or_none` replaced by `int`. Which makes sense, as `seconds_fraction_str` is `None`, setting `milliseconds` explicitely to zero in line 323.

[BTW I always get a bit confused by this "C and x or y" statements without brackets and have to look up Python's operator precedence. E.g. "x if C else y" would be faster to perceive and could be a better option for future changes.]

ebu_tt_live/bindings/_ebuttdt.py
nigelmegitt: Seems to work. We don't have a requirement for the 3 decimal places constraint so I just use a float in 50ce306.

ebu_tt_live/bindings/_ebuttdt.py
nigelmegitt: `_int_or_none` removed in 50ce306.

ebu_tt_live/bindings/test/test_times.py
nigelmegitt: The 3 decimal places constraint was not a spec requirement, and the new approach in 50ce306 means it is no longer needed. Nevertheless I have added tests for 3 decimal places and also in 50ce306 some tests for some 4 decimal place values.

ebu_tt_live/bindings/test/test_times.py
```diff
@@ -0,0 +1,92 @@
...
+        '1.003h': timedelta(hours = 1, seconds=10, milliseconds = 800),
+        '1.030h': timedelta(hours = 1, minutes = 1, seconds = 48),
+        '1.300h': timedelta(hours = 1, minutes = 18),
+        '1h': timedelta(hours = 1),
+        '1.3ms': timedelta(milliseconds = 1, microseconds = 300),
+        '1.03ms': timedelta(milliseconds = 1, microseconds = 30),
+        '1.30ms': timedelta(milliseconds = 1, microseconds = 300),
+        '1.003ms': timedelta(milliseconds = 1, microseconds = 3),
+        '1.030ms': timedelta(milliseconds = 1, microseconds = 30),
+        '1.300ms': timedelta(milliseconds = 1, microseconds = 300),        '1ms': timedelta(milliseconds = 1),
```
spoeschel: The 1ms test somehow is not on a separate line.

spoeschel:
Thank you for applying all the changes!

PR #509, opened 2019-10-10
Resequencer: fix sequence number 1 if no content

spoeschel:
While the EBUTT3Splicer is provided with the current sequence number
stored/incremented by the Resequencer, the `create_compatible_document`
method so far refers to the sequence number stored internally by
EBUTT3DocumentSequence instead. This leads to a sequence number of 1 in
the issued output document if no subtitle content is available.

To align with the correct (strictly monotonic increasing) sequence
number, always use the one stored by the Resequencer.

Fixes ebu/ebu-tt-live-toolkit#502.

### DIFF
--- a/ebu_tt_live/documents/ebutt3.py
+++ b/ebu_tt_live/documents/ebutt3.py
@@ -706,7 +706,7 @@ class EBUTT3DocumentSequence(TimelineUtilMixin, CloningDocumentSequence):
             clock_mode=self._reference_clock.clock_mode,
             sequence_identifier=self._sequence_identifier,
             authors_group_identifier=self.authors_group_identifier,
-            sequence_number=self._last_sequence_number,
+            sequence_number=kwargs.get('sequence_number', self._last_sequence_number),
             lang=self._lang
         )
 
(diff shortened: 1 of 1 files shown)


spoeschel:
Follow-up of #503.

ebu_tt_live/documents/ebutt3.py
```diff
@@ -706,7 +706,7 @@ def create_compatible_document(self, *args, **kwargs):
             clock_mode=self._reference_clock.clock_mode,
             sequence_identifier=self._sequence_identifier,
             authors_group_identifier=self.authors_group_identifier,
-            sequence_number=self._last_sequence_number,
+            sequence_number=kwargs.get('sequence_number', self._last_sequence_number),
```
nigelmegitt: Doesn't this _always_ use the specified sequence number if one was specified and the document is empty, even if there is a `_last_sequence_number`? I think that's the opposite of what we want, which is to use `_last_sequence_number` if it is `not None` or otherwise to try to use the specified `sequence_number` or if that fails, default to 1.

spoeschel: To prefer  `_last_sequence_number` would mean that it is always used, as the constructor initializes it to `0`. However the variable is [updated by received documents, if lower than the present value](https://github.com/ebu/ebu-tt-live-toolkit/blob/release/2.1.2/ebu_tt_live/documents/ebutt3.py#L916). So in case e.g. a segment length of 2 seconds was configured and every 0.5s a document was received, this would result in output sequence numbers of 4, 8, 12 etc. 

As I'm not aware of other possible use cases of `create_compatible_document`, my intention was to not touch any other cases here and therefore to use the supplied sequence number here (if present), as already the case if the sequence has any segments stored.

spoeschel:
I will also check for BDD tests here.

Issue #447, opened 2017-06-05
[COMPONENT] Resequencer not documented


nigelmegitt:
The Resequencer node should be documented in the "scripts and their functions page.

PR #510, opened 2019-10-10
Resequencer: allow to immediately issue documents

spoeschel:
So far the resequencer starts to regularly issue documents only after
the first EBU-TT Live document has been received. However for some use
cases this might be inconvenient as an active document must exist at all
times e.g. when creating segments for an MPEG-DASH stream. The reason
for documents being issued only after the first received EBU-TT Live
document is that certain parameters of that document are used for
initialisation.

This commit adds a new configuration parameter that specifies a document
which will be used for initialisation (instead of the first received
EBU-TT Live document). Therefore the resequencer will immediately (be
able to) start issuing documents after its creation.

Closes ebu/ebu-tt-live-toolkit#505.

### DIFF
--- a/docs/source/configurator.rst
+++ b/docs/source/configurator.rst
@@ -99,6 +99,7 @@ Node type dependent options for [nodeN] : ::
    ├─sequence_identifier : sequence identifier, default "re-sequencer"
    ├─segment_length : duration of each output segment in seconds, default 2
    ├─begin_output : ["immediate" (default) | {begin time} ] the time at which the first output segment should begin.
+   ├─init_document : File path to a document used to set initial parameters and afterwards start issuing documents (instead of awaiting the first received document), default None
    ├─discard : whether to discard content that has been encoded, default True
    └─clock
      └─type : ["local" (default) | "auto" | "clock"]
--- a/ebu_tt_live/node/consumer.py
+++ b/ebu_tt_live/node/consumer.py
@@ -76,7 +76,7 @@ class ReSequencer(AbstractProducerNode, SimpleConsumer):
     _expects = EBUTT3Document
     _provides = EBUTT3Document
 
-    def __init__(self, node_id, reference_clock, segment_length, discard, sequence_identifier,
+    def __init__(self, node_id, reference_clock, segment_length, init_document, discard, sequence_identifier,
                  consumer_carriage=None, producer_carriage=None, **kwargs):
         super(ReSequencer, self).__init__(
             node_id=node_id,
@@ -91,6 +91,20 @@ class ReSequencer(AbstractProducerNode, SimpleConsumer):
         self._segment_counter = 1
         self._sequence_identifier = sequence_identifier
         self._discard = discard
+        
+        if init_document is not None:
+            # Create sequence from init document, in order to immediately start document output
+            log.info('Creating document sequence from init document {}'.format(
+                init_document
+            ))
+            with open(init_document, 'r') as xml_file:
+                xml_content = xml_file.read()
+            xml_doc = EBUTT3Document.create_from_xml(xml_content)
+            
+            self._sequence = EBUTT3DocumentSequence.create_from_document(xml_doc, verbose=self._verbose)
+            if self._reference_clock is None:
+                self._reference_clock = self._sequence.reference_clock
+        
 
     @property
     def last_segment_end(self):
--- a/ebu_tt_live/node/consumer.py
+++ b/ebu_tt_live/node/consumer.py
@@ -76,8 +76,8 @@ class ReSequencer(AbstractProducerNode, SimpleConsumer):
     _expects = EBUTT3Document
     _provides = EBUTT3Document
 
-    def __init__(self, node_id, reference_clock, segment_length, init_document, discard, sequence_identifier,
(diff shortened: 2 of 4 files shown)


ebu_tt_live/node/consumer.py
```diff
@@ -76,7 +76,7 @@ class ReSequencer(AbstractProducerNode, SimpleConsumer):
     _expects = EBUTT3Document
     _provides = EBUTT3Document
 
-    def __init__(self, node_id, reference_clock, segment_length, discard, sequence_identifier,
+    def __init__(self, node_id, reference_clock, segment_length, init_document, discard, sequence_identifier,
```
nigelmegitt: Is `init_document` required? If not, we should set a default `=None` for it. It is not enough to rely on the default from the config read by `ebu-run`, because this class can be instantiated by a script separately from that config mechanism.

docs/source/configurator.rst
```diff
@@ -99,6 +99,7 @@ Node type dependent options for [nodeN] : ::
    ├─sequence_identifier : sequence identifier, default "re-sequencer"
    ├─segment_length : duration of each output segment in seconds, default 2
    ├─begin_output : ["immediate" (default) | {begin time} ] the time at which the first output segment should begin.
+   ├─init_document : File path to a document used to set initial parameters and afterwards start issuing documents (instead of awaiting the first received document), default None
```
nigelmegitt: It'd be good to add a section of documentation explaining how the init document is used and what constraints there are for it. For example, what happens if the init document has a different sequence id to the sequence we are connecting to? Does that mean that the expected incoming documents cause an error? (I expect it does but have not checked yet)

ebu_tt_live/node/consumer.py
```diff
@@ -91,6 +91,20 @@ def __init__(self, node_id, reference_clock, segment_length, discard, sequence_i
...
+        if init_document is not None:
+            # Create sequence from init document, in order to immediately start document output
+            log.info('Creating document sequence from init document {}'.format(
+                init_document
+            ))
+            with open(init_document, 'r') as xml_file:
+                xml_content = xml_file.read()
+            xml_doc = EBUTT3Document.create_from_xml(xml_content)
+            
+            self._sequence = EBUTT3DocumentSequence.create_from_document(xml_doc, verbose=self._verbose)
```
nigelmegitt: At this stage, we are duplicating code that is run by `process_document()` - did you consider simply calling `self.process_document(xml_doc)` instead?

docs/source/configurator.rst
spoeschel: Adding more explanation makes sense. What location within the docs would you prefer for such section? Ddocuments with a sequence ID different to the init document's one are discarded, raising an `IncompatibleSequenceError` with `ERR_DOCUMENT_NOT_COMPATIBLE`.

ebu_tt_live/node/consumer.py
spoeschel: You're right; I therefore added a default value in 9fbe83f37ac74ca4a888f7143a3fe3fabd681c4d.

ebu_tt_live/node/consumer.py
spoeschel: Actually I considered this, but the call to `EBUTT3DocumentSequence.add_document` then had the undesired consequence that the init document's sequence number is "burnt" i.e. cannot be used a second time by any real input document. Also the log messages would indicate that a real document had been received, instead of refering to the init document. So I duplicated/adjusted the relevant code part.

But in 564c6021d4f98a67188edbd1696fa7e66d56f81c I just moved the common parts into a new function to have a SPOF here.

spoeschel:
I haven't checked with the BDD tests yet, but I will do so.

ebu_tt_live/node/consumer.py
nigelmegitt: Ah, that's really interesting, I hadn't spotted there's a code path for setting up the sequence without setting any expectation in code about sequence number. This looks good, thank you.

docs/source/configurator.rst
nigelmegitt: Oh dear, I've found #447 as a result of this. We never documented the resequencer properly! We should have added it to `docs/source/scripts_and_their_functions.rst`. I'll open a quick pull request for that issue; if you could add some extra text to that pull request, that would be helpful. Otherwise I think this is now good to merge.

nigelmegitt: OK I've opened #510 to address #447, so if you review that and we merge it @spoeschel then you can merge that pull request here too and then add the documentation to the new section introduced there, or we can do it the other way round, merge this and then add to the Resequencer documentation. I don't mind which.

spoeschel: In this case I would prefer to first have this MR being merged. I then would add a commit to #515, adding documentation for this MR, and having that MR being merged, too. This saves us an additional issue/MR process.

spoeschel:
After rebasing on `release/2.1.2` I extended the resequencer documentation (which was recently added by #515).

docs/source/configurator.rst
nigelmegitt: Sorry I did this the other way around. I lost track a bit!

spoeschel: No problem!

PR #512, opened 2019-10-10
Python3 migration

### LINKED ISSUE #488: Move to Python 3

nigelmegitt:
This pull request closes #488.

Each fix was broken down in a separate commit with a (hopefully) informative message, however, we could squash all commits when merging to master. 

This pull request replaces #489, and travis should be able to build it successfully if it passes tests etc.


nigelmegitt:
OK we now have a working basis for a Python3 build based on some commits prior to the current `release/2.1.2` branch. As more commits get layered into the `release/2.1.2` branch we will need to also copy them into the `release/3.0` branch to keep it level. 

This is a bit painful; perhaps someone has a good pattern for managing this?

PR #513, opened 2019-10-11
Issue 499 bdd feature files

### LINKED ISSUE #499: Some BDD feature files aren't being run
The following files exist, but aren't being referenced in any `test_*.py` files, so they aren't being run.

features/styles/ebuttd_segmenter.feature
features/styles/fontSize_inherited.feature
features/styles/padding.feature

I was looking for a step definition in ebuttd_segmenter but couldn't find it!

nigelmegitt:
Add test steps to run feature files that were not previously run; closes #499.

### DIFF
--- a/testing/bdd/features/styles/padding.feature
+++ b/testing/bdd/features/styles/padding.feature
@@ -6,8 +6,8 @@ Feature: Compute padding on a single EBU-TT Live element
   # Percentage relative to width and height of region.
 
   Examples:
-  | xml_file    | cell_resolution | style_attribute | elem_id |  
-  | padding.xml | 10 10           | tts:padding     | region1 |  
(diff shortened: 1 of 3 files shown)


nigelmegitt:
Currently two tests are skipped in `padding.feature` because the code doesn't do what the test thinks it should do. @EyalLavi I think you wrote the feature a long long time ago. If you have any recollection of the intent, that might help us decide what to do about this.

This leaves the `ebuttd_segmenter.feature` still not running, but I think that may have been overtaken by events - there is no ebuttd segmenter per se; instead there is a resequencer and a deduplicator and an ebuttd encoder, and they each have tests (more to come for the ebuttd encoder) so my proposal is to delete the file. Again, @EyalLavi does that sound reasonable to you?

PR #516, opened 2019-10-16
Handle timing types in elements without crashing

### LINKED ISSUE #492: Error when parsing XML ebuttm:documentStartOfProgramme
I am trying to parse an EBU-TT-Live document using `EBUTT3Document.create_from_xml(xml)`. The `xml` variable is a string read in from a file.

The XML contains the following in the `<ebuttm:documentMetadata>` block:
```
<ebuttm:documentStartOfProgramme>10:00:00:00</ebuttm:documentStartOfProgramme>
```

However I'm getting a KeyError in `ebu_tt_live/bindings/_ebuttdt.py:73`: No such key `timing_attribute_name` in the XML parsing context.

nigelmegitt:
Fixes #492

* Define the type of `ebuttm:documentStartOfProgramme` correctly.
* When processing a timing type outside of the context of an attribute, bypass the timebase validation.

This is possibly a temporary hack, since element timing validation might be needed one day, but since we don't do anything with metadata elements, this isn't a disaster (yet).

### DIFF
--- a/ebu_tt_live/bindings/_ebuttdt.py
+++ b/ebu_tt_live/bindings/_ebuttdt.py
@@ -70,20 +70,25 @@ class _TimedeltaBindingMixin(object):
             # This means we are in XML parsing context. There should be a timeBase and a timing_attribute_name in the
             # context object.
             time_base = context['timeBase']
-            timing_att_name = context['timing_attribute_name']
-            if time_base not in cls._compatible_timebases[timing_att_name]:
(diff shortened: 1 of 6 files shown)


nigelmegitt:
Thanks for the review @spoeschel , since I would like to push ahead with a final Python2 release very soon (maybe today!) I'm going to merge this now. If I can do the other pull requests that are ok to merge then I will.

PR #518, opened 2019-10-18
Specify incompatible attributes

### LINKED ISSUE #514: EBUTT3DocumentSequence: format ERR_DOCUMENT_NOT_COMPATIBLE strings
In `EBUTT3DocumentSequence` the `ERR_DOCUMENT_NOT_COMPATIBLE` is used without specifying the actually affected attribute names [[1]](https://github.com/ebu/ebu-tt-live-toolkit/blob/release/2.1.2/ebu_tt_live/documents/ebutt3.py#L645) [[2]](https://github.com/ebu/ebu-tt-live-toolkit/blob/release/2.1.2/ebu_tt_live/documents/ebutt3.py#L650).

nigelmegitt:
Fixes #514.

### DIFF
--- a/ebu_tt_live/documents/ebutt3.py
+++ b/ebu_tt_live/documents/ebutt3.py
@@ -642,12 +642,12 @@ class EBUTT3DocumentSequence(TimelineUtilMixin, CloningDocumentSequence):
         if self.sequence_identifier != document.sequence_identifier or \
                 self._reference_clock.time_base != document.time_base:
             raise IncompatibleSequenceError(
-                ERR_DOCUMENT_NOT_COMPATIBLE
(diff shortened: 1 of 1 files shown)


nigelmegitt:
Thanks @spoeschel , to specify the attribute names I would have to refactor the code also: I want to do the minimum possible to get this minor bug fixed, so I'll go with this for now. If this is a significant issue that you can't tell if the incompatibility is with the sequence identifier or the timebase attribute, please could you raise that as a new issue?

spoeschel:
I'm sorry; this was a misunderstanding: I agree that the error message is sufficient to support users in further debugging. I just meant that the parameter `attr_names` should explicitely be specified in the `format` call, like this:

    ERR_DOCUMENT_NOT_COMPATIBLE.format(attr_names='ebuttp:sequenceIdentifier or ttp:timeBase')

PR #523, opened 2019-11-14
Improved EBU-TT-D encoding, including better tests and denester

nigelmegitt:
Add detailed conversion tests to improve the EBU-TT Live to EBU-TT-D conversion.

This also adds a Denester node that flattens nested div and span elements, which meets a requirement of EBU-TT-D.

This work represents contributions from a number of BBC engineers - thank you!


docs/source/scripts_and_their_functions.rst
```diff
@@ -107,6 +107,17 @@ attributes are duplicated.
 For the default configuration of the node, see:
 ``ebu-run --admin.conf=ebu_tt_live/examples/config/deduplicator_fs.json``
 
+Denester Node
+-------------
+This node flattens nested ``div`` and ``span`` elements such that no
+``div`` ends up containing a ``div`` and no ``span`` ends up containing
+a ``span``. It also removes any ``p`` elements that specify a ``region``
+attribute that differs from a specified region on an ancester element.
```
spoeschel: There is a small typo here ("ancester" instead of "ancestor")

nigelmegitt: Fixed in 5d0c24b

ebu_tt_live/config/node.py
```diff
@@ -192,6 +192,20 @@ def __init__(self, config, local_config):
         self._create_input(config)
         self._create_output(config)
 
+class Denester(ConsumerMixin, ProducerMixin, NodeBase):
```
spoeschel: The class has to be referenced further down in [nodes_by_type](https://github.com/ebu/ebu-tt-live-toolkit/blob/bbc-python3-staging/ebu_tt_live/config/node.py#L352), in order to be used e.g. in a config file.

ebu_tt_live/config/node.py
```diff
@@ -192,6 +192,20 @@ def __init__(self, config, local_config):
...
+    required_config = Namespace()
+
+    def _create_component(self, config):
+        self.component = processing_node.DenesterNode(
+            node_id=self.config.id
+        )
+
+    def __init__(self, config, local_config):
+        super(RetimingDelay, self).__init__(config, local_config)
```
spoeschel: This should be `Denester` instead of `RetimingDelay` ;-)

ebu_tt_live/node/denester.py
```diff
@@ -0,0 +1,332 @@
...
+            dataset["styles"] = document.binding.head.styling.style
+        else:
+            dataset["styles"] = []
+        dataset["document"] = document.binding
+        for div in divs:
+            unnested_divs.extend(DenesterNode.recurse(div,dataset))
+        unnested_divs = DenesterNode.combine_divs(unnested_divs)
+        unnested_divs = DenesterNode.check_p_regions(unnested_divs)
+        document.binding.body.div = unnested_divs
+        return document
```
spoeschel: Testing the Denester, no output documents are generated at all here. As far as I can see, returning the document is not enough but instead `self.producer_carriage.emit_data` has to be called to forward the processed document.

docs/source/configurator.rst
```diff
@@ -125,6 +125,9 @@ Node type dependent options for [nodeN] : ::
 
    type="distributor" : No options
 
+   type="denester"
+   └─sequence_identifier : sequence identifier, default "Denester1"
```
spoeschel: This option is not (yet) supported by the Denester.

ebu_tt_live/config/node.py
nigelmegitt: Agreed, I think we have some updates to make to this pull request based on further work, and this is one of them.

ebu_tt_live/config/node.py
nigelmegitt: Yikes! Good spot, thanks for that. Weird python, I wonder what it actually does when we write it like this?!

ebu_tt_live/node/denester.py
nigelmegitt: Agreed, I see that I have some changes made on the BBC fork that include a fix for that, so I will push an update to this pull request.

docs/source/configurator.rst
nigelmegitt: Again, to be fixed by pushing more commits here - we have this fixed in the BBC fork. Apologies!

ebu_tt_live/config/node.py
spoeschel: Python then aborts with a type error:

    TypeError: super(type, obj): obj must be an instance or subtype of type

nigelmegitt: Oh I see, fair enough. Weird that we never hit that bug.

docs/source/configurator.rst
spoeschel: OK, no problem! So I will put my review on hold, until I get news from you regarding this MR.

PR #525, opened 2019-11-15
Make region required in Part 3 docs

### LINKED ISSUE #524: layout element is wrongly allowed to be empty
Tech3350 v1.2 specifies that there must be at least 1 `region` element in a `layout` element, and in Part 1, that the `layout` element must be present. Tech3370 relaxes the constraint that `layout` must be present, but _not_ the constraint that if a `layout` element exists it must have at least 1 `region` child. Currently the XSD and code wrongly allows empty `layout` elements.

nigelmegitt:
Updates the XSD to match the spec, which requires that if a `layout` element is present, it must have at least one `region` child. In Part 3 the `layout` element itself is optional, but in Part 1 it is required.

Also update the test XML templates to be valid: many of them had invalid empty `layout` elements.

Fix the Segmenter so that if a `layout` element has no `region` children, the `layout` element itself is removed.

Closes #524.


ebu_tt_live/documents/ebutt3_segmentation.py
```diff
@@ -158,6 +158,7 @@ def _prune_orphan_elements(self):
             self._segment.head.styling = None
         if len(self._segment.head.layout.region) == 0:
             self._segment.head.layout.region = None
```
spoeschel: Is this line still needed?

nigelmegitt: hmm that looks strange, like it should say `self._segment.head.layout = None` instead. Good catch.

spoeschel: Actually the newly added line just below (line 161) contains the call you mention. The question is whether the current line (160) is still needed, as the new line (161) affects the parent node of line 160's node.

nigelmegitt: Oh yes I see what you mean, it does look like that line can be removed. I'll test removing it and if all goes well I'll push that change. Thank you @spoeschel !

nigelmegitt: Done in 5011a16b3dc790c571234b92391da9712c70a064

PR #536, opened 2020-09-17
Fix nltk dependency

### LINKED ISSUE #533: 'invalid syntax' error from nltk
I was getting the following error when running `ebu-dummy-encoder`:
```
Traceback (most recent call last):
  File "/opt/env/bin/ebu-dummy-encoder", line 11, in <module>
    load_entry_point('ebu-tt-live', 'console_scripts', 'ebu-dummy-encoder')()
  File "/opt/env/lib/python2.7/site-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/opt/env/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2852, in load_entry_point
    return ep.load()
  File "/opt/env/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2443, in load
    return self.resolve()
  File "/opt/env/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2449, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/opt/ebu_tt_live/scripts/ebu_dummy_encoder.py", line 6, in <module>
    from ebu_tt_live.documents.ebutt3 import EBUTT3Document
  File "/opt/ebu_tt_live/documents/__init__.py", line 1, in <module>
    from .base import SubtitleDocument, TimeBase, DocumentSequence
  File "/opt/ebu_tt_live/documents/base.py", line 2, in <module>
    from ebu_tt_live.utils import ComparableMixin
  File "/opt/ebu_tt_live/utils.py", line 9, in <module>
    from nltk import BlanklineTokenizer, PunktSentenceTokenizer, WhitespaceTokenizer
  File "/opt/env/lib/python2.7/site-packages/nltk/__init__.py", line 128, in <module>
    from nltk.collocations import *
  File "/opt/env/lib/python2.7/site-packages/nltk/collocations.py", line 35, in <module>
    from nltk.probability import FreqDist
  File "/opt/env/lib/python2.7/site-packages/nltk/probability.py", line 333
    print("%*s" % (width, samples[i]), end=" ")
                                          ^

SyntaxError: invalid syntax
```

Turns out my version of nltk (3.5) doesn't work with Python 2.7. So in `setup.py` I had to add an upperbound:
```
    install_requires=[
        ...
        "nltk<3.5",
        ...
    ]
```

spoeschel:
Use the last version that supports Python 2.7.

Fixes #533.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ sphinx-autobuild
 sphinxcontrib-plantuml
 recommonmark
 sphinx-rtd-theme
-pytest-bdd
+pytest-bdd<3.3.0
 pytest-cov
 pytest-mock
 pytest-twisted
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ setup(
         "service_identity",
         "twisted",
         "autobahn<18",
-        "nltk",
+        "nltk<3.5",
         "sortedcontainers",
         "configman",
         "six",


nigelmegitt:
@spoeschel the build fails, but not obviously because of the change to a specified version of nltk. I think we may need to resolve some other issue first, but I'd be happy to do it in the same pull request. What do you think?

spoeschel:
With version 3.2.1 of Pytest-BDD, the error doesn't occur and the (non-skipped) tests all pass. That version is actually the same one that automatically was used by the most recent *passing* Travis builds.

The two errors that show up with a recent version are caused by the comments that are present at the line ends of some feature example tables ([example](https://github.com/ebu/ebu-tt-live-toolkit/blob/master/testing/bdd/features/styles/style_attribute_simple.feature#L24)). While moving the comments to a separate line fixes the errors, the tests still don't pass. Hence this seems to be a larger issue.

So I think the easiest solution would be:
- add an upper bound (`<3.3.0`) for Pytest-BDD as well, for the Python 2 branch (as part of this MR is fine for me)
- open a separate issue in order to address the test errors with Pytest-BDD 3.3.0+ for the Python 3 branch in the future

Would this be OK for you?

nigelmegitt:
Ah, that problem. We've already fixed that in the BBC python3 fork. Yes, happy to pin the Pytest-BDD version for the python2 branch to make a release that works.

nigelmegitt (approved):
Thanks @spoeschel this looks good to me. I changed the target branch to release/2.1.3 following the normal procedure of creating a release staging branch. When we want to make a release we'll pull the changes from release/2.1.3 into master and create the release tag.

Feel free to merge this!

spoeschel:
Thanks, @nigelmegitt! I will have a look at possible further maintenance fixes for a 2.1.3 release.

PR #539, opened 2020-09-21
Pin used Python 2 packages

### LINKED ISSUE #538: Pin used packages for Python 2 maintenance branch
As a more generic and future-proof fix for issues like #533, add all currently used package versions to `requirements.txt` using `pip freeze`.

spoeschel:
Prevents issues like #533 in the future, as it is expected that further
dependencies will remove Python 2 support in the future.

The updated file has been generated using `pip freeze` and replacing the
repository path of the toolkit by `.`. The `pkg-resources==0.0.0` line
has been removed, as it was present due to [a bug](https://github.com/pypa/pip/issues/4022).

The `--index-url` parameter has been removed, as the specified URL
meanwhile forwards to the default package index.

Closes #538.


nigelmegitt (approved):
Thanks, builds and tests successfully, this is what we need for a stable release that will keep working for as long as the specified versions of upstream dependencies remain available.

PR #565, opened 2023-10-11
build: Added Docker build alternative

CCMA-Enginyeria:
Hi!
We've added an alternative to build/run using Docker and have updated the README.md to reflect this change. If you need more details, please let us know.
Thank you!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM python:2.7
+RUN apt-get update
+RUN apt-get install nodejs npm -y
+RUN mkdir /app
+WORKDIR /app
+COPY ./ /app
+RUN pip install virtualenv
+RUN virtualenv env
+RUN make
+ENTRYPOINT /bin/bash
--- a/README.md
+++ b/README.md
@@ -72,6 +72,17 @@ After that the following command should work:
 
     ebu-dummy-encoder
 
+Docker User
+=============
+
+Create image:
+
+    docker build -t ebu-tt-live-toolkit .
+
+Run container:
+
+    docker run -it ebu-tt-live-toolkit
+
 The Schema definitions XSD
 ==========================
 


nigelmegitt:
Hi, thanks for this. Please could you explain why this is needed?

CCMA-Enginyeria:
Thank you for your fast reply @nigelmegitt .
We believe that Docker could offer us a faster and simpler way to use the toolkit, as there's no need to worry about the OS or any dependencies; you only need to have the Docker runtime installed. (We've tested it on a Windows host.)

nigelmegitt:
Thanks @CCMA-Enginyeria - and why do you think this is needed in the main repo? I prefer to minimise the external dependencies as much as possible, and it looks as though this is something you could put into your own fork if you want it.

By the way, moving up to Python3.x is highly desirable for the project, and significant work was done in the BBC fork at https://github.com/bbc/ebu-tt-live-toolkit - I mention this in case it makes your build environment easier to manage.

CCMA-Enginyeria:
Ok, understood. If you prefer not to add more dependencies, it’s better not to merge that PR. 
If anybody asks you, we’ll have Docker support on our fork: https://github.com/CCMA-Enginyeria/ebu-tt-live-toolkit 😉. 
We’ll check the BBC fork; it seems they’ve done a great job. Thank you, @nigelmegitt.

nigelmegitt:
Closing due to lack of motivation for the change. If there's a strong community desire or need for a docker-based build in future I'd be happy to revisit this decision and consider adding docker support.

andreastai:
Thanks @CCMA-Enginyeria! I think a Docker build is very helpful for the user community, especially because the current version still runs under Python 2.7, which is phased out on most Systems. To have a Docker build was also discussed before.

I would be in favour of adding this to the main repo because it makes the current code usable. This is also a very "thin" code addition.

